PR #3, opened 2017-07-17
Implement circuit/relay/0.1.0 spec

vyzo:
Updates the implementation for the new v0.1.0 spec.

Some notable additions:
- Error codes are amended; see https://github.com/libp2p/specs/pull/24
- CAN_HOP is implemented; see https://github.com/libp2p/specs/pull/28


relay.go
```diff
@@ -52,166 +61,271 @@ func NewRelay(ctx context.Context, h host.Host, opts ...RelayOpt) (*Relay, error
...
 	if len(ctp) == 0 {
-		return &RelayStatus{
-			Code:    StatusRelayNotConnected,
-			Message: "refusing to make new connection for relay",
+		if !r.active {
+			r.handleError(s, pb.CircuitRelay_HOP_NO_CONN_TO_DST)
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(r.ctx, time.Second*10)
```
Stebalien: Should probably go in a constant (and 10s is a *long* time).

relay.go
```diff
@@ -52,166 +61,271 @@ func NewRelay(ctx context.Context, h host.Host, opts ...RelayOpt) (*Relay, error
...
+		defer cancel()
+		err = r.host.Connect(ctx, dst)
+		if err != nil {
+			log.Debugf("error opening relay connection to %s: %s", dst.ID.Pretty(), err.Error())
+			r.handleError(s, pb.CircuitRelay_HOP_CANT_DIAL_DST)
+			return
```
Stebalien: Again, we don't really need to explicitly connect. We can just do the following:

1. If `len(ctp) == 0 && !r.active`, error and return.
2. Create a timeout context.
3. Try to open stream (will automatically open the conn if necessary).

listen.go
```diff
@@ -26,20 +28,20 @@ func (r *Relay) Matches(a ma.Multiaddr) bool {
...
-		log.Infof("accepted relay connection: %s", c.ID())
-		s := RelayStatus{
-			Code:    StatusOK,
-			Message: "OK",
-		}
-		if err := s.WriteTo(c); err != nil {
+		err := l.Relay().writeResponse(c.Stream, pb.CircuitRelay_SUCCESS)
+		if err != nil {
+			log.Debugf("error writing relay response: %s", err.Error())
+			c.Stream.Close()
```
vyzo: But we do want the other side to be able to continue to write, don't we?
This is a transparent proxy in effect, the other side's writes will continue to be forwarded by the second goroutine until it also closes the channel.

util.go
```diff
@@ -2,62 +2,89 @@ package relay
...
+	peer "github.com/libp2p/go-libp2p-peer"
+	pstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
+	mh "github.com/multiformats/go-multihash"
 )
 
-type singleByteReader struct {
-	r io.Reader
-}
+func peerToPeerInfo(p *pb.CircuitRelay_Peer) (empty pstore.PeerInfo, err error) {
```
vyzo: I will modify to avoid using the named arguments, @whyrusleeping doesn't like them either I think.

util.go
```diff
@@ -2,62 +2,89 @@ package relay
...
-	l, err := binary.ReadUvarint(&singleByteReader{r})
-	if err != nil {
-		return nil, err
-	}
+type delimitedReader struct {
+	r   io.Reader
+	buf []byte
+}
+
+// the gogo protobuf NewDelimitedReader is buffered, which may eat up stream data
```
vyzo: Well, the only slowdown is in reading the length, where it could take multiple single byte reads. 
But this only happens during handshake, so it's a drop in the water in the connection's lifetime.

I will add a comment to that effect.

relay.go
vyzo: ok, will constant it and make it 1s or something.

relay.go
```diff
@@ -52,166 +61,271 @@ func NewRelay(ctx context.Context, h host.Host, opts ...RelayOpt) (*Relay, error
...
-func (r *Relay) Dial(ctx context.Context, relay peer.ID, dest ma.Multiaddr) (*Conn, error) {
-	s, err := r.host.NewStream(ctx, relay, HopID)
+func (r *Relay) Dial(ctx context.Context, relay pstore.PeerInfo, dest pstore.PeerInfo) (*Conn, error) {
+	err := r.host.Connect(ctx, relay)
 	if err != nil {
 		return nil, err
 	}
 
-	if err := writeLpMultiaddr(s, dest); err != nil {
+	s, err := r.host.NewStream(ctx, relay.ID, ProtoID)
```
vyzo: I was not aware that opening a new stream implicitly connects. Is that a recent development?

relay.go
vyzo: See above for unawareness of implicit connect on new stream construction. 
On the other hand, we want to distinguish Dial errors from Stream opening errors (there are separate error codes and helps debugging), so I think we want to keep the explicit connect.

listen.go
vyzo: Actually, I misunderstood the context of the code, and gave a confusing reply.

vyzo: The proxy is not in effect yet, so my comment about the transparent proxying is largely irrelevant.

You are right that we do want to close the output side of the channel in some way, but on the other hand the other side is not doing any writing at this point -- it is waiting for a response per protocol.
I will add a comment here.

relay.go
Stebalien: Not really. It's actually necessary to prevent some potential race conditions (open a connection, connection gets garbage collected (once we start actually doing that), try to open a stream). It's also just really convenient.

relay.go
Stebalien: Ah. Sorry, I just looked at the log messages, not the error codes. We might be able to look at the error itself but relying on error values in go is... :sob:.

util.go
Stebalien: Got it, I thought we were using this for the entire connection.

relay.go
vyzo: I will remove the Connect from the Dial method implementation as well -- and it won't need a peer info for the relay any more.

relay.go
Stebalien: Actually, we should probably ask @whyrusleeping about this. I'm not familiar with how long these things usually take in IPFS and this *may* have to lookup peer addresses in the DHT (which could take up to a minute...).

relay.go
vyzo: I am rewiring the logic a bit to use the test you suggested and share the timeout context with the stream opening. 
I am keeping the explicit connect here however for distinguishing the error conditions.

vyzo: Actually, we can interpret the error codes a bit more liberally, and use HOP_CANT_DIAL_DEST for the stream opening, and HOP_CANT_OPEN_DST_STREAM for handshake errors.

I am dropping the explicit connect after all.

relay.go
vyzo: I think 10s is maybe not so long time after all -- I suspect the current 2s might be too short.

I will set it back to 10s and let's wait to hear @whyrusleeping on this.

vyzo:
oh wow, a travis failure.

vyzo:
In terms of basic functionality, I think it's complete.
Next step is to integrate with libp2p using the transport interface, but this will be implemented in go-libp2p.

Stebalien:
LGTM

vyzo:
Added support for integrating relay as a transport to a host, through the `AddRelayTransport` function.

Depends on these PRs to work:
- https://github.com/multiformats/go-multiaddr/pull/53
- https://github.com/libp2p/go-addr-util/pull/4

so the tests will fail and we can't merge until these two are merged and imported to gx (and the package dependencies updated).

transport.go
```diff
@@ -4,12 +4,42 @@ import (
...
 	"fmt"
 
+	addrutil "github.com/libp2p/go-addr-util"
 	host "github.com/libp2p/go-libp2p-host"
 	swarm "github.com/libp2p/go-libp2p-swarm"
 	tpt "github.com/libp2p/go-libp2p-transport"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+const P_CIRCUIT = 290
```
Stebalien: I believe this constant needs to be added to the multicodec repo: https://github.com/multiformats/multicodec/

vyzo: It's already in the multiaddr repo (https://github.com/multiformats/multiaddr), do we actually need it in multicoded too?

vyzo: PR: https://github.com/multiformats/multicodec/pull/52

vyzo:
Alright, all green. I  am going to go ahead and merge it, we are ready for interop tests with js.

PR #4, opened 2017-07-24
Relay Discovery and unspecific address dialing

vyzo:
Implements passive relay discovery and unspecific relay address dialing, as discussed in https://github.com/libp2p/go-libp2p/issues/209. 

Implementation Notes:
- Relay discovery is implemented through a Notifee, which tests CanHop for all new connections and tracks a list of known hop relays.
- Unspecific address dialing is implemented by iterating through the list of known relays and attempting to dial through them; if none of the connection attempts succeed it fails with a RelayError.

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -3,6 +3,7 @@ package relay
 import (
 	"context"
 	"fmt"
+	"math/rand"
 
 	pb "github.com/libp2p/go-libp2p-circuit/pb"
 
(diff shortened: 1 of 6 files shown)


vyzo:
summoning @Stebalien @whyrusleeping

dial.go
```diff
@@ -42,19 +46,57 @@ func (d *RelayDialer) DialContext(ctx context.Context, a ma.Multiaddr) (tpt.Conn
...
 
 	return d.Relay().DialPeer(ctx, *rinfo, *dinfo)
 }
 
+func (d *RelayDialer) tryDialRelays(ctx context.Context, dinfo pstore.PeerInfo) (tpt.Conn, error) {
+	var relays []peer.ID
+	d.mx.Lock()
+	for p := range d.relays {
+		relays = append(relays, p)
+	}
```
Stebalien: FYI, the fast and idiomatic way to do this is:

```go
relays := append([]peer.ID{}, d.relays...)
```

Stebalien: Nevermind, relays isn't a slice.

Stebalien:
LGTM. Eventually, we'll probably want to use the DHT to find relays (or maybe even just gossip them) but asking connected peers is a good start.

---

Slightly off topic but it would be nice if peers could advertise us service lists on connect instead of having to test for everything (lots of extra round trips).

PR #7, opened 2017-07-29
Revert "Report partial circuit address in Conn.RemoteMultiaddr"

vyzo:
Reverts libp2p/go-libp2p-circuit#6
Per discussion with @Stebalien we concluded that the origianl behaviour is correct.


vyzo:
summoning @Stebalien

PR #17, opened 2017-09-13
update go-stream-muxer

Stebalien:
Use the new Reset method.


relay.go
```diff
@@ -146,12 +144,14 @@ func (r *Relay) CanHop(ctx context.Context, id peer.ID) (bool, error) {
...
 	err = wr.WriteMsg(&msg)
 	if err != nil {
+		s.Reset()
 		return false, err
 	}
-
 	msg.Reset()
 
 	err = rd.ReadMsg(&msg)
+	s.Close()
```
vyzo: there is a defer s.Close(), this isn't necessary.

vyzo: ah, you removed the defered close.

relay.go
```diff
@@ -258,52 +258,54 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
...
 	log.Infof("relaying connection between %s and %s", src.ID.Pretty(), dst.ID.Pretty())
 
+	// Don't reset streams after finishing or the other side will get an
+	// error, not an EOF.
 	go func() {
 		count, err := io.Copy(s, bs)
 		if err != io.EOF && err != nil {
 			log.Debugf("relay copy error: %s", err)
 		}
-		s.Close()
```
vyzo: Small nitpick -- is there a reason for the reordering?
I'd like to have all the i/o related functionality grouped together, before the log message.

relay.go
```diff
@@ -258,52 +258,54 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
...
 		log.Debugf("relayed %d bytes from %s to %s", count, dst.ID.Pretty(), src.ID.Pretty())
+		s.Close()
 	}()
 
 	go func() {
 		count, err := io.Copy(bs, s)
 		if err != io.EOF && err != nil {
 			log.Debugf("relay copy error: %s", err)
 		}
-		bs.Close()
```
vyzo: same here.

relay.go
Stebalien: No reason, I originally moved this to a defer but then decided against it and put it back in the wrong place. I'll put it back.

relay.go
Stebalien: (I didn't want to reset on error and then close).

relay.go
vyzo: ok, it does read nicer after the reordering!

vyzo:
hrm, I think we missed the canHop error handler -- we need a reset there too.
I'll open a new quick one.

Stebalien:
Where?

vyzo:
#19

PR #19, opened 2017-09-13
handleCanHop: reset stream on errors

vyzo:


Stebalien:
Good catch!

Stebalien:
Meh. They pass on linux.

PR #20, opened 2017-10-08
tag 'can-hop' peers to aid in connection closing logic

whyrusleeping:
This should also probably tag relay peers with a higher value when they are actually relaying a connection for us. The logic there isnt simple, so i've punted on it for now.

### DIFF
--- a/notify.go
+++ b/notify.go
@@ -47,6 +47,7 @@ func (n *RelayNotifiee) Connected(s inet.Network, c inet.Conn) {
 			n.mx.Lock()
 			n.relays[id] = struct{}{}
 			n.mx.Unlock()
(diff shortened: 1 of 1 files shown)


vyzo:
this will tag all connections to relays effectively, is that what we want?

whyrusleeping:
@vyzo unsure, I think we want *at least* this. Connections to peers that
will relay our connections are more valuable than otherwise. I think we
also want to mark connections that are currently relaying for us as even
more valuable, but tracking that seems tricky

vyzo:
we should certainly tag connections that are actively relayed with high priority.
what does the value of 2 mean priority-wise?

whyrusleeping:
They're somewhat arbitrary right now, I'm using 5 for dht kbucket peers.

PR #29, opened 2018-03-08
Correctly handle stream errors

Stebalien:
If copying from one stream to another fails, *reset* the streams. Otherwise, we propagate an EOF and the other side thinks the connection was gracefully terminated.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -293,19 +293,29 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	// error, not an EOF.
 	go func() {
 		count, err := io.Copy(s, bs)
-		if err != io.EOF && err != nil {
+		if err != nil {
 			log.Debugf("relay copy error: %s", err)
+			// Reset both.
+			s.Reset()
+			bs.Reset()
+		} else {
+			// propagate the close
+			s.Close()
 		}
-		s.Close()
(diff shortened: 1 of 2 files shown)


relay.go
```diff
@@ -293,19 +293,29 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	// error, not an EOF.
 	go func() {
 		count, err := io.Copy(s, bs)
-		if err != io.EOF && err != nil {
+		if err != nil {
```
vyzo: so this will treat `io.EOF` as an error and reset. 
Is this really what we want? 
Or is it that `Copy` doesn't error with `EOF` at all?

dryajov:
Is this addressing https://github.com/ipfs/interop/issues/16?

relay.go
Stebalien: Copy doesn't return EOFs (it considers that a success case):

> A successful Copy returns err == nil, not err == EOF. Because Copy is defined to read from src until EOF, it does not treat an EOF from Read as an error to be reported.

vyzo: Cool, that's definitely an improvement.

Stebalien:
@dryajov it wasn't intended to. This is just a drive-by fix. Does it fix that issue?

dryajov:
It doesn't seem like it would.

PR #30, opened 2018-03-11
fix: dont error out on unknown addrs in peer info

dryajov:
This PR tries to fix the `browser-browser-go` case reported in https://github.com/ipfs/interop/issues/16. The js circuit will send all the addrs it knows of the src and dst peers, but go will error out if it doesn't know one, since we have addrs that are valid only in some implementations, it causes this error. In this case, the websocket start addr was being sent (code 479).

### DIFF
--- a/util.go
+++ b/util.go
@@ -29,9 +29,8 @@ func peerToPeerInfo(p *pb.CircuitRelay_Peer) (pstore.PeerInfo, error) {
 	for i := 0; i < len(addrs); i++ {
 		a, err := ma.NewMultiaddrBytes(p.Addrs[i])
 		if err != nil {
-			return pstore.PeerInfo{}, err
+			addrs[i] = a
 		}
-		addrs[i] = a
 	}
 
 	return pstore.PeerInfo{ID: peer.ID(h), Addrs: addrs}, nil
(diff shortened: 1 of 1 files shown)


util.go
```diff
@@ -29,9 +29,8 @@ func peerToPeerInfo(p *pb.CircuitRelay_Peer) (pstore.PeerInfo, error) {
 	for i := 0; i < len(addrs); i++ {
 		a, err := ma.NewMultiaddrBytes(p.Addrs[i])
 		if err != nil {
```
Stebalien: err == nil

vyzo:
@diasdavid indeed a bug -- but not a protocol issue :)

util.go
```diff
@@ -28,10 +28,9 @@ func peerToPeerInfo(p *pb.CircuitRelay_Peer) (pstore.PeerInfo, error) {
 	addrs := make([]ma.Multiaddr, len(p.Addrs))
 	for i := 0; i < len(addrs); i++ {
 		a, err := ma.NewMultiaddrBytes(p.Addrs[i])
-		if err != nil {
-			return pstore.PeerInfo{}, err
+		if err == nil {
+			addrs[i] = a
```
Kubuxu: This will cause the `addrs` to have gaps between addresses. Are you 100% sure it is ok?

Stebalien: Good catch. I believe so but we still shouldn't do this.

PR #33, opened 2018-03-13
Update for transport refactor

Stebalien:
DO NOT MERGE (yet)
Part of: libp2p/go-libp2p#297

(tests failing as gx deps need updating)


transport.go
```diff
@@ -22,29 +21,6 @@ var Protocol = ma.Protocol{
...
-	if err != nil {
-		panic(err)
-	}
-
-	err = addrutil.AddTransport(proto)
-	if err != nil {
-		panic(err)
-	}
-
-	for _, tp := range tps {
```
vyzo: is there equivalent code for this or is it handled in a different manner now?

Stebalien: It's handled by the swarm internally. See: https://github.com/libp2p/go-addr-util/pull/11

vyzo: is the swarm explicitly aware of relay addresses or does it ask the transports dynamically to respond whether they support the address?
If so, how are encapsulations handled?

Stebalien: The swarm asks its transports.

For listening, we just figure out which transport is responsible using the algorithm described [here](https://github.com/libp2p/go-libp2p-transport/blob/88e91d2b047d46b35012678dfa23b711d2ba5f8f/transport.go#L81-L94) and ask it to listen on the address.

For dialing, we lookup the appropriate transport (same algorithm) and call the `CanDial` function on it ([code](https://github.com/libp2p/go-libp2p-swarm/blob/feat/refactor/swarm_dial.go#L256)). We could just try dialing but this way allows us to filter out addresses we definitely can't dial higher up the stack.

vyzo: Not clear it will have the same behaviour -- for example we want to pick up relay for `/ip4/X.X.X.X/tcp/YY/p2p-circuit/ip4/X.X.X.X/tcp/YY/ipfs/Qmfoo` (that's explicit relay dialing).
What will the algorithm pick here?

Stebalien: I don't see how that's related to *this* function but it will pick the relay transport. It'll:

1. Iterate over the protocols (ip4, tcp, ipfs, p2p-circuit, ip4, tcp, ipfs).
2. When it gets to the p2p-circuit protocol, it will notice that the associated transport is a `Proxy` transport and will pass the address off to that transport.

vyzo: ok, being a proxy is the key feature.

transport.go
```diff
@@ -59,40 +35,45 @@ func (r *Relay) Transport() *RelayTransport {
 	return (*RelayTransport)(r)
 }
 
-func (t *RelayTransport) Dialer(laddr ma.Multiaddr, opts ...tpt.DialOpt) (tpt.Dialer, error) {
-	if !t.Matches(laddr) {
+func (t *RelayTransport) Listen(laddr ma.Multiaddr) (tpt.Listener, error) {
+	protos := laddr.Protocols()
+	if len(protos) == 0 || protos[len(protos)-1].Code == P_CIRCUIT {
```
vyzo: I am confused by this -- you fail with "not a relay address" if it's the passive relay address?

Stebalien: Interesting... So, I intended to write `!=` however, that *wouldn't* have worked because we listen with `/p2p-circuit/ipfs/QmId`. Really, it should support both.

Stebalien: Note: I wanted the check to be a bit more specific than "has a p2p-circuit protocol". However, I'll have to consider carefully how to do this properly.

relay.go
```diff
@@ -54,8 +57,9 @@ func (e RelayError) Error() string {
 	return fmt.Sprintf("error opening relay circuit: %s (%d)", pb.CircuitRelay_Status_name[int32(e.Code)], e.Code)
 }
 
-func NewRelay(ctx context.Context, h host.Host, opts ...RelayOpt) (*Relay, error) {
+func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ...RelayOpt) (*Relay, error) {
```
vyzo: where does one get a transport upgrader?

Stebalien: We construct it when constructing the transports.

https://github.com/libp2p/go-libp2p/blob/29228f250fa6b70ad45cfef02386353a8c36c0f9/config/config.go#L105-L143

PR #39, opened 2018-06-05
add in hop tracking

whyrusleeping:
Its sometimes nice to know how many connections you are relaying for other peers.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -107,12 +107,12 @@ func (r *Relay) rmLiveHop(from, to peer.ID) {
 	r.lhLk.Lock()
 	defer r.lhLk.Unlock()
 
-	r.lhCount--
 	trg, ok := r.liveHops[from]
 	if !ok {
 		return
 	}
 
+	r.lhCount--
 	delete(trg, to)
 	if len(trg) == 0 {
 		delete(r.liveHops, from)
--- a/relay.go
+++ b/relay.go
@@ -40,6 +40,10 @@ type Relay struct {
 
 	relays map[peer.ID]struct{}
 	mx     sync.Mutex
+
+	liveHops map[peer.ID]map[peer.ID]struct{}
+	lhCount  uint64
+	lhLk     sync.Mutex
 }
 
 type RelayOpt int
@@ -65,6 +69,7 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 		self:     h.ID(),
 		incoming: make(chan *Conn),
 		relays:   make(map[peer.ID]struct{}),
+		liveHops: make(map[peer.ID]map[peer.ID]struct{}),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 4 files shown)


relay.go
```diff
@@ -61,6 +65,7 @@ func NewRelay(ctx context.Context, h host.Host, opts ...RelayOpt) (*Relay, error
 		self:     h.ID(),
 		incoming: make(chan *Conn),
 		relays:   make(map[peer.ID]struct{}),
+		liveHops: make(map[peer.ID]map[peer.ID]struct{}),
```
Stebalien: FYI, you can do this with a single map and no counter:

```go
type hop struct {
  from peer.ID
  to peer.ID
}

// ...

liveHops: make(map[hop]struct{})
```

whyrusleeping: also with double the memory complexity

whyrusleeping: technically, i'm not using the maps here. But I would like to have that info exposed somehow

Stebalien: Damn go has efficient maps. I was expecting the extra map to have a lot of memory overhead. Digression: The memory complexity depends on how many nodes we have and how many hops we have per node. For a sane number of nodes (<=1000), the two map method is more efficient assuming 3+ relays per node. >= 10000 nodes, the two map method is less efficient up to ~40 relays per node.

> technically, i'm not using the maps here. But I would like to have that info exposed somehow

(technically, you *can* still do that with the one map method, it's just harder to lookup info about a specific node)

---

For fun:

```go
package main

import (
	"fmt"
	"runtime"
)

type hop struct {
	from, to string
}

func main() {
	things := make([]string, 10000)
	for i := range things {
		things[i] = fmt.Sprintf("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx%d", i)
	}
	n := 37
	maps(things, n)
	nomaps(things, n)
}

func maps(things []string, n int) {
	var before, after runtime.MemStats

	runtime.ReadMemStats(&before)

	m := make(map[string]map[string]struct{})

	for _, f := range things {
		mm := make(map[string]struct{})
		for _, t := range things[:n] {
			mm[t] = struct{}{}
		}
		m[f] = mm
	}
	runtime.ReadMemStats(&after)
	fmt.Println(after.Alloc - before.Alloc)
}

func nomaps(things []string, n int) {
	var before, after runtime.MemStats

	runtime.ReadMemStats(&before)

	m := make(map[hop]struct{})

	for _, f := range things {
		for _, t := range things[:n] {
			m[hop{f, t}] = struct{}{}
		}
	}
	runtime.ReadMemStats(&after)
	fmt.Println(after.Alloc - before.Alloc)
}
```

bigs:
fixed a logical error and rebased

Stebalien:
@bigs YES! :heart:

relay.go
```diff
@@ -84,6 +89,43 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 	return r, nil
 }
 
+func (r *Relay) addLiveHop(from, to peer.ID) {
+	r.lhLk.Lock()
+	defer r.lhLk.Unlock()
+
+	r.lhCount++
```
raulk: Should we be increasing this count only if we weren't tracking this hop already, i.e. (below)

```
if _, ok := tgt[to]; !ok {
  tgt[to] := struct{}{}
  r.lhCount++
}
```

Actually, let me check if the implementation guarantees non-duplicate relays, because if that's the case, we wouldn't get here to begin with.

Stebalien: You're right, we don't detect duplicates (which makes *all* this logic incorrect). We should either:

1. Track duplicates in this map (i.e., count connections).
2. Forbid duplicates.

I'd prefer the latter but that may lead to a race condition closing one connection and opening another (the relay may see us trying to keep two connections open. This may not be an issue but (1) will definitely be simpler.

PR #46, opened 2018-09-05
Fix counting logic to avoid double counting

### LINKED ISSUE #45: Fix hop tracking
Per discussion in #39, let's make some fixes to this tracking logic.

bigs:
Closes #45

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -41,7 +41,7 @@ type Relay struct {
 	relays map[peer.ID]struct{}
 	mx     sync.Mutex
 
-	liveHops map[peer.ID]map[peer.ID]struct{}
+	liveHops map[peer.ID]map[peer.ID]int
 	lhCount  uint64
 	lhLk     sync.Mutex
 }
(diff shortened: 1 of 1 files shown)


bigs:
sorry for the delay @Stebalien , chgd

relay.go
```diff
@@ -93,14 +93,19 @@ func (r *Relay) addLiveHop(from, to peer.ID) {
...
 	if !ok {
-		trg = make(map[peer.ID]struct{})
+		trg = make(map[peer.ID]int)
 		r.liveHops[from] = trg
 	}
+	var count int
+	if count, ok = trg[to]; !ok {
+		count = 0
+	}
+	count++
```
Stebalien: We can replace all of this with `trg[to]++`. That will:

1. Lookup `to` in `trg`. If it's not there, it'll return 0.
2. Increment.
3. Set `trg[to]`.

bigs: ahhh i was actually looking into these zero values and having a hard time finding out what was up. happy to change!

PR #47, opened 2018-09-25
Standalone usage of /p2p-circuit addr (dependency gardening)

lgierth:
Removal of the old constant is going to be a separate step.


lgierth:
This doesn't make much sense with gx's package concept

PR #48, opened 2018-09-26
Change transport address to not include /p2p part

lgierth:
Doesn't work yet, dialing via circuit breaks when the change to `RelayListener.Multiaddr()` is included.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -35,12 +35,14 @@ func (c *Conn) RemoteAddr() net.Addr {
 	}
 }
 
+// TODO: is it okay to cast c.Conn().RemotePeer() into a multiaddr? might be "user input"
 func (c *Conn) RemoteMultiaddr() ma.Multiaddr {
-	a, err := ma.NewMultiaddr(fmt.Sprintf("/ipfs/%s/p2p-circuit/ipfs/%s", c.Conn().RemotePeer().Pretty(), c.remote.ID.Pretty()))
-	if err != nil {
-		panic(err)
-	}
-	return a
+	proto := ma.ProtocolWithCode(ma.P_P2P).Name
+	peerid := c.Conn().RemotePeer().Pretty()
+	p2paddr := ma.StringCast(fmt.Sprintf("/%s/%s", proto, peerid))
+
+	circaddr := ma.Cast(ma.CodeToVarint(P_CIRCUIT))
+	return p2paddr.Encapsulate(circaddr)
 }
 
 func (c *Conn) LocalMultiaddr() ma.Multiaddr {
--- a/listen.go
+++ b/listen.go
@@ -1,7 +1,6 @@
 package relay
 
 import (
-	"fmt"
 	"net"
 
 	pb "github.com/libp2p/go-libp2p-circuit/pb"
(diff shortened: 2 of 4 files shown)


conn.go
```diff
@@ -35,12 +35,14 @@ func (c *Conn) RemoteAddr() net.Addr {
...
 }
 
+// TODO: is it okay to cast c.Conn().RemotePeer() into a multiaddr? might be "user input"
 func (c *Conn) RemoteMultiaddr() ma.Multiaddr {
-	a, err := ma.NewMultiaddr(fmt.Sprintf("/ipfs/%s/p2p-circuit/ipfs/%s", c.Conn().RemotePeer().Pretty(), c.remote.ID.Pretty()))
-	if err != nil {
-		panic(err)
-	}
-	return a
+	proto := ma.ProtocolWithCode(ma.P_P2P).Name
```
vyzo: Why `P_P2P`?

listen.go
```diff
@@ -50,6 +50,8 @@ func (l *RelayListener) Addr() net.Addr {
 }
 
 func (l *RelayListener) Multiaddr() ma.Multiaddr {
+	// TODO(lgierth): t0182 "default failure"
+	return ma.Cast(ma.CodeToVarint(P_CIRCUIT))
```
vyzo: wait, isn't there dead code below?

lgierth: Yeah I was trying to pin down the test failures -- this PR is by no means ready for merge yet

conn.go
lgierth: We're moving /ipfs to /p2p :)

lgierth:
This is ready now, and depends on libp2p/go-libp2p-swarm#87

conn.go
```diff
@@ -35,12 +35,14 @@ func (c *Conn) RemoteAddr() net.Addr {
...
+	peerid := c.Conn().RemotePeer().Pretty()
+	p2paddr := ma.StringCast(fmt.Sprintf("/%s/%s", proto, peerid))
+
+	circaddr := ma.Cast(ma.CodeToVarint(P_CIRCUIT))
+	return p2paddr.Encapsulate(circaddr)
```
vyzo: Isn't this a little inconsistent? The addr we get can't be combined with an ID to be dialable -- instead it would need to be parsed.

vyzo: More of a nit though, we _need_ the peer part to differentiate connections; otherwise we'd just report `/p2p-circuit`.
But maybe that's what we want with the new semantics?

lgierth: > Isn't this a little inconsistent? The addr we get can't be combined with an ID to be dialable -- instead it would need to be parsed.

Not sure what you mean -- you'll get `/p2p/QmRelay/p2p-circuit`, and you can do the following to get a full peer address:
```go
peeraddr := ma.StringCast("/p2p/QmFoo")
conn.RemoteMultiaddr().Encapsulate(peeraddr)
```

> More of a nit though, we need the peer part to differentiate connections; otherwise we'd just report /p2p-circuit.  But maybe that's what we want with the new semantics?

Yeah I think we want to stop treating the underlying transport connection as something that's unique to a peer. (cc @Stebalien)

vyzo: Ok, I read that as `/p2p/QmRelay/p2p-circult/p2p/QmPeer`.
Getting back just the relay + p2p-circuit as transport is fine and consistent.

PR #53, opened 2018-12-18
Don't log raw binary

anacrolix:
Causes issues particularly when logging to the terminal.

### DIFF
--- a/listen.go
+++ b/listen.go
@@ -33,7 +33,7 @@ func (l *RelayListener) Accept() (manet.Conn, error) {
 		}
 
 		// TODO: Pretty print.
-		log.Infof("accepted relay connection: %s", c)
+		log.Infof("accepted relay connection: %q", c)
 
 		return c, nil
 	case <-l.ctx.Done():


listen.go
```diff
@@ -4,7 +4,6 @@ import (
 	"net"
 
 	pb "github.com/libp2p/go-libp2p-circuit/pb"
-
```
vyzo: don't remove this line!

listen.go
```diff
@@ -33,7 +32,7 @@ func (l *RelayListener) Accept() (manet.Conn, error) {
 		}
 
 		// TODO: Pretty print.
-		log.Infof("accepted relay connection: %s", c)
+		log.Infof("accepted relay connection: %q", c)
```
vyzo: is this the right format specifier?
godoc says:
```
%q	a single-quoted character literal safely escaped with Go syntax.
```

anacrolix: It's the best I've found in this kind of circumstance (unknown strings).

vyzo: ok, let's go with it.

Issue #18, opened 2017-09-13
It should be possible to close a stream and then read from it.


Stebalien:
Putting [this](https://github.com/libp2p/go-libp2p-circuit/blob/d1fca568cd8d5d9e97fbd374b4b9e28028a9cd73/relay.go#L154) close call before the read causes the tests to fail. However, given that we're now allow half-closed streams, it should work.

It looks like something is seeing that the stream is closed and resetting the connection before responding.

vyzo:
So is this just a subtlety in the code? 
Semantically, the close should be after the read, but perhaps we should add a comment there.

Stebalien:
IMO, the close should come after the last write. That is, close means "I'm done sending requests".

Ichbinjoe:
This no longer seems to be the case as of now.

https://github.com/Ichbinjoe/go-libp2p-circuit/blob/preemptive-close/relay.go#L217

I made these changes and the tests appear to still pass. Whatever issue exists no longer seems to now.

Stebalien:
@Ichbinjoe mind making a PR? We'll need to make sure it doesn't break js-libp2p but it would be nice to get this fix in.

Ichbinjoe:
I currently don't have the JS stack setup, but I'm sure over the next couple of days I can mess around with it more.

jsign:
While doing a toy-app for an article, I wanted to be somewhat strict with the documentation.
If I navigate to the documented semantics of close, we can find [this](https://github.com/libp2p/go-libp2p-core/blob/master/mux/mux.go#L22).
> // Close closes the stream for writing. Reading will still work (that is, the remote side can still write).

In the toy-app case, the side that opened the stream would only read stuff, so guided by that comment, I switched a `defer`ed `Close()`, to an immediate `Close()` (before reading the stream), which lead to a `stream reset` error.

Am I talking about the same thing discussed here? I'm running just `go-libptp@v0.4.2` (asking since this isn't that repo). If that's the case, maybe it would be better to delete that comment since it may mislead to a usage that doesn't behave as expected?

PR #55, opened 2018-12-27
Preemptively close stream in CanHop

Ichbinjoe:
Since we are done writing, we can close the write side of the stream
before reading the resulting status.

Closes https://github.com/libp2p/go-libp2p-circuit/issues/18

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -214,13 +214,18 @@ func (r *Relay) CanHop(ctx context.Context, id peer.ID) (bool, error) {
 		return false, err
 	}
 
+	if err := wr.Close(); err != nil {
+		s.Reset()
+		return false, err
+	}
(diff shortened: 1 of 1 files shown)


vyzo:
does this add anything useful in terms of behaviour?

vyzo:
Also, we need to make sure this doesn't break js.

Ichbinjoe:
The only behavioral change this may bring is that our side will appear to hang up marginally sooner than previously.

You can see the discussion more in #18

Stebalien:
Se still need to test this against JS, just in case. JS may close the entire stream prematurely when it receives a EOF on the read side.

Stebalien:
ping @vasco-santos

PR #69, opened 2019-04-07
use io.CopyBuffer with explicitly allocated buffers

vyzo:
`io.Copy` uses a 32KB buffer, which as pprof indicates results in significant memory usage.
This changes to using `io.CopyBuffer` directly with 4KB buffers.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -376,7 +376,7 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	go func() {
 		defer r.rmLiveHop(src.ID, dst.ID)
 
-		count, err := io.Copy(s, bs)
+		count, err := io.CopyBuffer(s, bs, make([]byte, 4096))
 		if err != nil {
 			log.Debugf("relay copy error: %s", err)
 			// Reset both.
@@ -390,7 +390,7 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	}()
 
 	go func() {
-		count, err := io.Copy(bs, s)
+		count, err := io.CopyBuffer(bs, s, make([]byte, 4096))
 		if err != nil {
 			log.Debugf("relay copy error: %s", err)
 			// Reset both.
--- a/relay.go
+++ b/relay.go
@@ -41,8 +41,6 @@ type Relay struct {
 
 	incoming chan *Conn
 
-	bufPool pool.BufferPool
-
 	relays map[peer.ID]struct{}
 	mx     sync.Mutex
 
@@ -379,8 +377,8 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	go func() {
 		defer r.rmLiveHop(src.ID, dst.ID)
 
-		buf := r.bufPool.Get(4096)
-		defer r.bufPool.Put(buf)
+		buf := pool.Get(4096)
+		defer pool.Put(buf)
 
 		count, err := io.CopyBuffer(s, bs, buf)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


vyzo:
See https://github.com/golang/go/blob/master/src/io/io.go#L391

vyzo:
Yeah, we could use a sync.Pool, steb made the same comment. Does it ever return memory? I am worried about hoarding memory when we scale up.

vyzo:
reading the docs, it explicitly mentioned that pooled objects may be gc'ed which is nice.

raulk:
Yeah, it holds weak references, so unused elements are elegible for collection by the GC :-)

vyzo:
Added a pool.

relay.go
```diff
@@ -390,7 +400,10 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	}()
 
 	go func() {
-		count, err := io.Copy(bs, s)
+		buf := r.bufPool.Get().([]byte)
+		defer r.bufPool.Put(buf)
```
raulk: Can we zero out the buffer before putting it back? It might not be necessary in practice, but would rather do it so getting a buffer from the pool always returns a zero value byte slice, whether new or pooled.

vyzo: I don't think it's necessary, as it will contain encrypted data. Also, zeroing out takes time, I'd rather keep it as lightweight as possible.

vyzo: Also, the way `CopyBuffer` is implemented it doesn't look at the data. At worst we might leak some old encrypted data to the reader, which a) isn't happening in our stack and b) it is just random junk thanks to the encryption.

raulk: I think we should do it. Lets not leak data even if it’s encrypted. Leaking data would also corrupt the SecIO/TLS ciphers. Zeroing an array is very cheap.

vyzo: sigh.... I think it's completely unnecessary, but ok.

vyzo: There is no utility function to do that in go, and I really hate to have to write bzero.
Also, it's an O(n) operation.

vyzo: Also, how on earth are we going to leak data with it? 
The reader just puts bytes in the slice and returns the number of bytes read.
The writer then writes only the bytes that have been read.

vyzo: Also of note that we started with the intention of making it more efficient than allocating a new buffer. If we add zeroing out we pretty much lost all efficiency gains.

raulk: > The reader just puts bytes in the slice and returns the number of bytes read.
> The writer then writes only the bytes that have been read.

Yeah, that's right.

> Also of note that we started with the intention of making it more efficient than allocating a new buffer. If we add zeroing out we pretty much lost all efficiency gains.

We *are* more efficient this way -- wrt. to memory churn, which is what we're trying to optimise here. Zeroing in a 4096-byte slice on my machine takes 70ns/op, it's negligible, so I think it doesn't hurt us to reset an array to avoid leaking data on a pool. It's simply more correct, and removes a potential source of bugs.

vyzo: > It's simply more correct

it's not, the buffer contents are irrelevant in its pattern of use.

> removes a potential source of bugs.

I don't think we remove any source of bugs here, we are just adding redundant (ugly) code!

raulk: Still, it's proper housekeeping. But let's see what @Stebalien thinks!

raulk: @Stebalien thinks it's unnecessary, so I yield, @vyzo! Good discussion.

Stebalien: Note: If we _want_ to do this, we could try adding this feature to the buffer pool library directly (given some good real-world benchmarking). However, this is probably one of the less important places to clear the buffer given that we're handling encrypted data for a party that doesn't really trust us anyways.

relay.go
```diff
@@ -85,6 +87,11 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 		incoming: make(chan *Conn),
 		relays:   make(map[peer.ID]struct{}),
 		liveHops: make(map[peer.ID]map[peer.ID]int),
+		bufPool: sync.Pool{
+			New: func() interface{} {
+				return make([]byte, 4096)
```
Stebalien: So, welcome to go. Converting a `[]byte` to an `interface{}` _allocates_. Let's just use https://github.com/libp2p/go-buffer-pool (https://godoc.org/github.com/libp2p/go-buffer-pool#Get)

vyzo: Oh wow, that's nuts.
Yeah, I'll use go-buffer-pool.

PR #70, opened 2019-04-10
Use NoDial option when opening hop streams for non-active relays

vyzo:
It turns out that the relays are doing dialing on NewStream in the wild, presumably because they still have dead connections.
This uses the new NoDial option to instruct the swarm to not dial the hop target instead of explicitly checking for open connections and is robust to failed connections.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ env:
   global:
     - GOTFLAGS="-race"
   matrix:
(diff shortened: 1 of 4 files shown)


vyzo:
gx build failures are to be expected...

vyzo:
@Stebalien LGTM'ed over slack, as he is on mobile, so merging.

PR #71, opened 2019-04-11
use buffer pool in newDelimitedReader

vyzo:
instead of allocating the buffer flat, given how short-lived it is.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -323,6 +323,7 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 	// stop handshake
 	rd := newDelimitedReader(bs, maxMessageSize)
 	wr := newDelimitedWriter(bs)
+	defer rd.Close()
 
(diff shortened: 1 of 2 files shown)


vyzo:
Actually there is no useful place to put the buffer back in the pool.

vyzo:
I added a `Close` method in the reader to release the buffer.

vyzo:
The global pool probably works just as well for this, and they can be reused by the hop stream buffers.

vyzo:
Oh you mean a pool for the actual readers. Yeah, maybe so.

Stebalien:
Yep.

vyzo:
I will merge as is, as using a pool for the delimitedReaders themselves adds quite a bit of ugly complexity (we need to track the buffers in two steps).

Stebalien:
Well, we could just leave the buffers inside the `delimitedReader` pool. But either way.

PR #74, opened 2019-05-05
Hard Limit the number of hop stream goroutines

vyzo:
This adds a hard limit to the number of hop goroutines, so that relays don't get overloaded.

Note that the live hop tracking has been removed for two reasons:
- it was not by any way exported to the outside world, just taking memory
- there is lock contention

Note 2: DO NOT MERGE AS IS; I will rebase/squash before merging to clean up history.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	pb "github.com/libp2p/go-libp2p-circuit/pb"
@@ -29,6 +30,9 @@ var (
 	RelayAcceptTimeout   = 10 * time.Second
 	HopConnectTimeout    = 30 * time.Second
 	StopHandshakeTimeout = 1 * time.Minute
+
+	HopStreamBuffer = 4096
+	HopStreamLimit  = 1 << 19 // 512K hops for 1M goroutines
 )
 
 // Relay is the relay transport and service.
@@ -47,9 +51,9 @@ type Relay struct {
 	relays map[peer.ID]struct{}
 	mx     sync.Mutex
 
-	liveHops map[peer.ID]map[peer.ID]int
-	lhCount  uint64
-	lhLk     sync.Mutex
+	// atomic counters
+	sCount  int32
+	lhCount int32
 }
 
 // RelayOpts are options for configuring the relay transport.
@@ -88,7 +92,6 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 		self:     h.ID(),
 		incoming: make(chan *Conn),
 		relays:   make(map[peer.ID]struct{}),
-		liveHops: make(map[peer.ID]map[peer.ID]int),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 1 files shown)


vyzo:
An alternative to hard resetting is to add a new error code for overloaded relays.
But if we do this, we must stop lingering awaiting EOF in `handleError`; perhaps we should just reset after sending the error.

vyzo:
Resetting the stream won't work, the error will not be propagated further.
But we _can_ reduce the EOF timeout to something much more reasonable than the current 1 minute (to say 5s).
Fortunately this is a variable in `go-libp2p-net`, so the consumer can set it.

vyzo:
Implemented the `RELAY_OVERLOADED` error code, so that we don't do hard resets without notifying the other side. 
The relay daemon can set the `EOFTimeout` to something shorter in order to reduce the goroutine linger time.

vyzo:
I reverted to resetting the stream when the hop limit is exceeded -- being nice has deleterious effects in the number of lingering goroutines.

relay.go
```diff
@@ -47,9 +51,9 @@ type Relay struct {
 	relays map[peer.ID]struct{}
 	mx     sync.Mutex
 
-	liveHops map[peer.ID]map[peer.ID]int
-	lhCount  uint64
-	lhLk     sync.Mutex
+	// atomic counters
+	sCount  int32
```
Stebalien: Nit: could we give these full names? (`streamCount`?)

relay.go
```diff
@@ -29,6 +30,9 @@ var (
 	RelayAcceptTimeout   = 10 * time.Second
 	HopConnectTimeout    = 30 * time.Second
 	StopHandshakeTimeout = 1 * time.Minute
+
+	HopStreamBuffer = 4096
```
Stebalien: ultra nit: `HopStreamBufferSize`

pb/relay.proto
```diff
@@ -21,6 +21,7 @@ message CircuitRelay {
     STOP_DST_MULTIADDR_INVALID = 351;
     STOP_RELAY_REFUSED         = 390;
     MALFORMED_MESSAGE          = 400;
+    RELAY_OVERLOADED           = 500;
```
Stebalien: Are we going to use this or should we just leave it at a reset?

relay.go
```diff
@@ -29,6 +30,9 @@ var (
...
+	HopStreamLimit  = 1 << 18 // 256K hops for 512K goroutines
```
Stebalien: Have we checked this against our current numbers? This seems kind of low, actually. For 20k peers, this'll give us less than 10 streams per peer.

relay.go
vyzo: Sure, will do.

relay.go
vyzo: ok!

pb/relay.proto
vyzo: I will drop it in the rebase/squash.

relay.go
vyzo: We can easily double it -- in fact the mplex relay where I am testing this is running with double the count (it overrides in the daemon init).

vyzo:
> Sounds like something we need although I'm a bit worried we should be using per-peer limits instead. At the moment, ~500 peers could fully mesh-connect through the relay to kill it.

Per-peer limits are a little complex to implement, and need a lock (which I would like to avoid).
We have not observed fully connected meshes in the relays whatsoever; what has been observed is a long tail distribution where most peers have a small number of streams (they are connecting to the peers behind the relay) and then a bunch have a large number of streams (they are accepting connections).

relay.go
vyzo: I'll try with quadruple the count (hop limit at 1M) and evaluate memory usage with it.

vyzo: Seems like we are tight on memory with 2M goroutines active.

vyzo: I doubled the default, and actual daemons can set it higher if they have the resources.

relay.go
vyzo: done.

relay.go
vyzo: done

Stebalien:
> Per-peer limits are a little complex to implement, and need a lock (which I would like to avoid).
We have not observed fully connected meshes in the relays whatsoever; what has been observed is a long tail distribution where most peers have a small number of streams (they are connecting to the peers behind the relay) and then a bunch have a large number of streams (they are accepting connections).

I'm primarily worried about someone attacking a relay this way but this certainly isn't the only way.

vyzo:
rebased/squashed to just 2 commits and dropped the `RELAY_OVERLOADED` change in pb.

PR #75, opened 2019-05-07
Tag peers with live hop streams

### LINKED ISSUE #65: Work with the connection manager
We should integrate with the connection manager to deprioritize relay-hop connections when under heavy load.

vyzo:
This adds connection manager tagging for peers with live hops.
Note the extra supervisory goroutine in the hop, as we need to ensure both sides are closed before removing the tag and reduce the live hop count.

Closes #65

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -118,10 +118,15 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 
 func (r *Relay) addLiveHop(from, to peer.ID) {
 	atomic.AddInt32(&r.liveHopCount, 1)
+	r.host.ConnManager().UpsertTag(from, "relay-hop-stream", func(v int) int { return v + 1 })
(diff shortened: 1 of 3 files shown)


vyzo:
Reusing the existing goroutine will liner too, and will make decrementing the stream count awkward. 
Plus the new goroutine has a very short stack, which is not necessarily the case when reusing the goroutine.

But I do like the idea of the atomic counter, it's more code but totally obviates the need for the extra goroutine.

vyzo:
Alright, implemented with an atomic counter.

relay.go
```diff
@@ -364,10 +369,18 @@ func (r *Relay) handleHopStream(s inet.Stream, msg *pb.CircuitRelay) {
 
 	r.addLiveHop(src.ID, dst.ID)
 
+	goroutines := new(int32)
```
vyzo: note that this is allocated on the heap to avoid lingering the original goroutine stack.

PR #76, opened 2019-05-08
call Stream.Reset instead of Stream.Close

Stebalien:
May fix https://github.com/ipfs/go-ipfs/issues/6237

Basically,

1. We hang while closing a stream (because `Close` waits). Also, https://github.com/libp2p/go-mplex/issues/9
2. This blocks the connection manager because it assumes that close _doesn't_ wait.

This may also fix a stream leak.


vyzo:
tests fail...

Stebalien:
Yeah, we expect a nice EOF.

vyzo:
we need the `Shutdown` call already.

Stebalien:
1. We need `Close` to mean "close both ways" (with a sane default timeout), `Reset` to mean "throw everything away", and then yeah, `CloseWrite` and `CloseRead`.

vyzo:
The test failure seems insurmountable, we'll need to change the stream API for this to work :(

Stebalien:
Not necessarily. The test failure is due to the assumption that `conn.Close()` flushes. However, `conn.Close()` actually _resets_ in our TCP transport as well (for efficiency) so this isn't really an issue.

Stebalien:
But I agree this sucks.

conn.go
```diff
@@ -28,29 +29,53 @@ func (n *NetAddr) String() string {
 	return fmt.Sprintf("relay[%s-%s]", n.Remote, n.Relay)
 }
 
+func (c *Conn) Close() error {
+	return c.stream.Reset()
```
hsanjuan: Maybe this is relevant here. This is how I ended up doing it in gostream :

https://github.com/hsanjuan/go-libp2p-gostream/blob/master/conn.go#L38

```go
// Close closes the connection.
// Any blocked Read or Write operations will be unblocked and return errors.
func (c *conn) Close() error {
	if err := c.s.Close(); err != nil {
		c.s.Reset()
		return err
	}
	go pnet.AwaitEOF(c.s)
	return nil
}
```

If I remember well, bluntly resetting libp2p streams on `Close()` caused errors on the other side on situations where the stream closing was supposed to be a clean operation.

Stebalien: Yeah, but we need to be careful:

1. Concurrent reads aren't safe. We need to interrupt any concurrent reads (somehow) before we can do this. We _should_ be able to do this by setting a read deadline but... we'll have to make sure to fix any/all of our multiplexers/transports (deadlines currently don't interrupt).
2. We probably want to set some timeouts/deadlines.
3. We should probably call `Close()` in the background as well. That will mimic a normal `close(tcpFileDescriptor)` call (which will let the kernel flush in the background).

IMO, the _real_ issue here is that `TCPConn.Close` is actually a very fancy best-effort function that does a bunch of dirty work in the background. Ideally, even if we make `Close` close both directions, we wouldn't be _that_ sloppy.

Stebalien:
@vyzo, this is getting really critical and _both_ the TCP and the QUIC transports reset the underlying connection on close without flushing anything (I specifically modified the TCP transport to do this to avoid unnecessary work). This bug likely just crashed our gateways and has been causing problems for a ton of people.

The only other solution I can think of is to:

1. Close the stream. We'll have to fix any bugs we have where closing streams doesn't necessarily interrupt any in-progress writes (an issue for both yamux and mplex).
2. Set a read deadline on the stream. https://github.com/whyrusleeping/yamux/pull/28 fixes read deadlines for yamux but they're still broken in mplex.
3. Wait for the current reader to exit (read lock).
4. Read on the stream waiting for an EOF or a timeout.
5. On timeout, reset.
6. Do all this in a goroutine. so we don't stall anything.

But that requires fixing multiple annoying bugs.

vyzo:
Sigh... If it has been identified as _the_ critical bug, then we got to merge it.

Stebalien:
> Sigh... If it has been identified as the critical bug, then we got to merge it.

Well, Close is stalling. I'll try shipping a patch to some of our partners to make sure this _is_ the bug.

vyzo:
This needs a rebase for the go mod debacle (go.sum conflicts)

vyzo:
rebased and resolved the conflict; should be ready for merge.

vyzo:
we also need the companion in #77, so that the connection manager doesn't kill the underlying hop relay connections when they have stop streams.

conn.go
```diff
@@ -28,29 +29,53 @@ func (n *NetAddr) String() string {
...
+func (c *Conn) SetDeadline(t time.Time) error {
+	return c.stream.SetDeadline(t)
+}
+
+func (c *Conn) SetReadDeadline(t time.Time) error {
+	return c.stream.SetReadDeadline(t)
+}
+
+func (c *Conn) SetWriteDeadline(t time.Time) error {
+	return c.stream.SetReadDeadline(t)
```
raulk: Wrong method.

conn.go
```diff
@@ -11,7 +12,7 @@ import (
 )
 
 type Conn struct {
-	inet.Stream
+	stream inet.Stream
```
raulk: I don't think you need to remove the embedding. You should be able to override the `Close()` method delegating the call to `c.Stream.Close()`.

conn.go
vyzo: wrong method, spotted by raulk; will fix.

vyzo: fixed

conn.go
vyzo: This is also narrowing the interface to that of a `Conn`, so it's not the same.

PR #77, opened 2019-05-22
Tag the hop relay when creating stop streams

vyzo:
This tags the underlying hop relay with the connection manager when we open a new stop stream.
This should avoid the comical situation of dropping the hop relay connection and killing all the relayed connections with it in the connection manager.
On top of #76

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"time"
 
+	host "github.com/libp2p/go-libp2p-host"
 	inet "github.com/libp2p/go-libp2p-net"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
@@ -14,7 +15,7 @@ import (
 type Conn struct {
 	stream inet.Stream
 	remote pstore.PeerInfo
-	relay  *Relay
+	host   host.Host
 }
 
 type NetAddr struct {
@@ -67,13 +68,13 @@ func (c *Conn) RemoteAddr() net.Addr {
 // by the connection manager, taking with them all the relayed connections (that may themselves
 // be protected).
 func (c *Conn) tagHop() {
-	c.relay.host.ConnManager().UpsertTag(c.stream.Conn().RemotePeer(), "relay-hop-stream", incrementTag)
+	c.host.ConnManager().UpsertTag(c.stream.Conn().RemotePeer(), "relay-hop-stream", incrementTag)
 }
 
 // Decrement the underlying relay connection tag by 1; this is performed when we close the
 // relayed connection.
 func (c *Conn) untagHop() {
-	c.relay.host.ConnManager().UpsertTag(c.stream.Conn().RemotePeer(), "relay-hop-stream", decrementTag)
+	c.host.ConnManager().UpsertTag(c.stream.Conn().RemotePeer(), "relay-hop-stream", decrementTag)
 }
 
 // TODO: is it okay to cast c.Conn().RemotePeer() into a multiaddr? might be "user input"
--- a/conn.go
+++ b/conn.go
@@ -14,6 +14,7 @@ import (
 type Conn struct {
 	stream inet.Stream
 	remote pstore.PeerInfo
+	relay  *Relay
 }
 
 type NetAddr struct {
@@ -30,6 +31,7 @@ func (n *NetAddr) String() string {
 }
 
 func (c *Conn) Close() error {
+	c.untagHop()
 	return c.stream.Reset()
 }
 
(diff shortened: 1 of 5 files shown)


conn.go
```diff
@@ -28,29 +30,62 @@ func (n *NetAddr) String() string {
...
+func (c *Conn) SetDeadline(t time.Time) error {
+	return c.stream.SetDeadline(t)
+}
+
+func (c *Conn) SetReadDeadline(t time.Time) error {
+	return c.stream.SetReadDeadline(t)
+}
+
+func (c *Conn) SetWriteDeadline(t time.Time) error {
+	return c.stream.SetReadDeadline(t)
```
raulk: Wrong method.

conn.go
```diff
@@ -11,8 +12,9 @@ import (
 )
 
 type Conn struct {
-	inet.Stream
+	stream inet.Stream
```
raulk: I don't think you need to remove the embedding. You should be able to override the `Close()` method delegating the call to `c.Stream.Close()`.

conn.go
vyzo: Wrong pr! This is on the base pr, please review that on #76

conn.go
vyzo: again, wrong PR; these are on the base pr.

raulk:
@vyzo I changed the target branch of this PR to the correct one.

conn.go
raulk: Taken to #76.

vyzo:
I was planning to merge it on master separately because it solves a different issue, can we change the target branch back again? We can also do this after we merge #76.

relay.go
```diff
@@ -118,14 +118,14 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 
 func (r *Relay) addLiveHop(from, to peer.ID) {
 	atomic.AddInt32(&r.liveHopCount, 1)
-	r.host.ConnManager().UpsertTag(from, "relay-hop-stream", func(v int) int { return v + 1 })
-	r.host.ConnManager().UpsertTag(to, "relay-hop-stream", func(v int) int { return v + 1 })
+	r.host.ConnManager().UpsertTag(from, "relay-hop-stream", incrementTag)
```
raulk: Worth mentioning that these tags are only used on the hop relay node. And are used for bookkeeping the weight of the peers we're bridging, so when the connection manager prunes connections, it'll pick those peers that are relying less on us.

relay.go
```diff
@@ -118,14 +118,14 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
...
+	r.host.ConnManager().UpsertTag(to, "relay-hop-stream", incrementTag)
 }
 
 func (r *Relay) rmLiveHop(from, to peer.ID) {
 	atomic.AddInt32(&r.liveHopCount, -1)
-	r.host.ConnManager().UpsertTag(from, "relay-hop-stream", func(v int) int { return v - 1 })
-	r.host.ConnManager().UpsertTag(to, "relay-hop-stream", func(v int) int { return v - 1 })
+	r.host.ConnManager().UpsertTag(from, "relay-hop-stream", decrementTag)
```
raulk: Same comment as in `addLiveHop(from, to peer.ID)`; here we're maintaining those tags.

conn.go
```diff
@@ -60,6 +62,14 @@ func (c *Conn) RemoteAddr() net.Addr {
 	}
 }
 
+func (c *Conn) tagHop() {
+	c.relay.host.ConnManager().UpsertTag(c.stream.Conn().RemotePeer(), "relay-hop-stream", incrementTag)
```
raulk: In a `A --> R --> B` interaction, this `Conn` object represents `A --> R` and `B --> R`. For every connection we make to the relay, we are boosting the hop relay score, to tell the connection manager we are relying on that relay for connectivity to other peers.

conn.go
```diff
@@ -14,6 +14,7 @@ import (
 type Conn struct {
 	stream inet.Stream
 	remote pstore.PeerInfo
+	relay  *Relay
```
raulk: Unless we actually need the `Relay` here, let's replace this with `Host`.

P.S.: I would expect the Relay to have an accessor for the `peer.ID` of the relay, but it doesn't. That would've made the tagging code much clearer, as currently the reader needs to understand that the stream's remote ID is the relay, not the actual final peer.

relay.go
vyzo: ok

conn.go
vyzo: It made more sense to use the `Relay` object because it's the creator of those connections, but we can switch that to be a host.

PR #80, opened 2019-09-18
return the full address from conn.RemoteMultiaddr

Stebalien:
I can't think of a good reason not to do this (see https://github.com/libp2p/go-libp2p/issues/723). I thought there was one but I can't remember what it was.

This also changes how we parse addresses to use `ma.SplitFunc`, avoiding some allocations.


vyzo:
test failure:
```
panic: multiaddr failed to parse: no protocol with code 290
```

PR #86, opened 2019-11-16
Relay Hop Filters

Jorropo:
Closes libp2p/go-libp2p#741.

The goal is to have an list of peer id in the config of IPFS (`swarm.ExtraAllowedHop` (I didn't find any clear key name right now)) (or any other app using libp2p).
I also added an already implemented filter by peer id helpers with dynamic access to the map (`Allow`, `Unallow`, `IsAllowed`) to maybe allow adding this in the API for auto adding via ipfs-cluster, but I don't really expect this be used often.
The helper is mainly to example how an filter could be made.

Its partialy retro compatible.
Its if user was using `relay.Opt*`.
Its not if user was using numbers.


Stebalien:
So, instead of hard-coding this into the relay, what if the relay just accepted a filter function? That way, the user can use any logic it wants to filter connections:

* Filter by subnet.
* Filter by peer ID.
* Authenticate peers using some side-band protocol.

Note: we'll probably want to switch to the functional options pattern (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) in that case.

Jorropo:
@Stebalien done !

relay.go
```diff
@@ -95,18 +85,17 @@ func NewRelay(ctx context.Context, h host.Host, upgrader *tptu.Upgrader, opts ..
 		self:     h.ID(),
 		incoming: make(chan *Conn),
 		relays:   make(map[peer.ID]struct{}),
+		isAllowedToHop: func(_ network.Stream) bool {
+			return false
+		},
```
vyzo: This is not a reasonable default! The default should be to allow all nodes to hop.

relay.go
```diff
@@ -266,7 +255,7 @@ func (r *Relay) handleNewStream(s network.Stream) {
 }
 
 func (r *Relay) handleHopStream(s network.Stream, msg *pb.CircuitRelay) {
-	if !r.hop {
+	if !r.isAllowedToHop(s) {
```
vyzo: we need to add an error code specific to rejections (something like `HOP_REJECTED`) instead of confusing users trying to debug the issue.

relay.go
vyzo: This needs to be separate from the filter.

Jorropo:
@vyzo said:
> We need to separate `OptHop` from the filter, as they are two different things.
The `OptHop` options should continue working as it is today, and use the filter as a secondary mechanism to make a decision on whether to allow hopping or not.
The default filter should be to allow all when `OptHop` is enabled.

Keeping `OptHop` is for me a bad idea mainly because that very unclear, I first though of that but that will not help devlopers foreign to circuit :
`OptHop` means is hop allowed, can people hop on this node ?
So to properly implement that line relay.go#258 (in the patch) should be `if !(r.hop && r.isAllowedToHop(s)) {` but that repetive, we have 2 option to do the same thing and that source of confusion (isAllowedToHop can just return false if hop isn't be activated).
I can only find 2 advantage of doing thing like this :
- It remove the need for the user to keep a reference of what the previous filter before replacing it with `func(_ network.Stream) bool {return false}`, but `hop` and `isAllowedToHop` are private so its impossible to change them while execution.
- It allow to very easly know if the node should advertise it self as relay in the dht, but things are not done that way, no body read `hop` to know if we advertise the relay (I'm not sure if relay discovery is done via the dht or random walk).

@vyzo said:
> This is not a reasonable default! The default should be to allow all nodes to hop.

I completely agree with you but that not the current default and I want to keep this change as non breaking as possible, I will change that if more people agree with us.
Its better for network health to have relay enabled by default because there isn't a lot of peers relaying so the charge is bigger and bigger on relay as people stop relaying due to the cost it is, with relay enabled by default cost of relaying will getting lower and lower over time (reversing the snowball effect).

@vyzo said:
> We need to add an error code specific to rejections (something like HOP_REJECTED) instead of confusing users trying to debug the issue.

I think that useless and counter productive (and breaking change with other implementation of circuit), that needed only if we got rejected but we still need to know if we can tell other peer about the hopability of the peer wich just rejected us.
But I don't expect to use this as ban list because that will cost more than the attacker need to do to counter it (create a new identity) (and that will be a very simple dos attack, create an id, make your self banned and repeat, you will fill the memory of the targeted machine, could be done very very fastly if the peer id was pregenerated with a GPU).
And I expect user using filter to implement their own discovery mechanism (a config file for peer id filter, mdns for subnet filter or just send the peer id of the relay in the side band protocol).

So this is a case when a peer rejected you is still probably hopable by other random peer wich is impossible to reach in real world condition (except with a ban but in this case this filter is a very bad place to implement it, identify or the security layer would take that way better).

Performance is critical in libp2p, I don't think we can take a small part of it just to be polite in the debug log because that add up and at the end politeness slows down the network. (And I don't think user not aware will notice difference between `peer doesn't allow hopping` and `peer refused hopping` (and again the second sentence may mean the same as the first one)).

Jorropo:
@vyzo said:
> The OptHop options should continue working as it is today, and use the filter as a secondary mechanism to make a decision on whether to allow hopping or not.

`OptHop` is currently our mechanism to make decision to allow hopping or not.
Also `OptHop` will be not breaking, `Relay.hop` is.

relay.go
Jorropo: I completely agree with you but that not the current default and I want to keep this change as non breaking as possible, I will change that if more people agree with us.

relay.go
Jorropo: We don't need to create breaking change with other code base and older version just to be polite and print the right error message.

vyzo:
We don't break existing code just like that...

vyzo:
We probably also need to change the `CanHop` handler to check the filter.

Jorropo:
@vyzo wdym ? Strangely all my changes break nothing (if user was using OptHop and not number for config), but the type change was asked by @Stebalien, I first placed a type switch but this is cleaner.

Jorropo:
I tested circuit example and without any change it works as before.

vyzo:
You are deprecating `OptHop`; that's a no.

Jorropo:
Ah ok, I just renamed it `OptHopAll`, its not really its just more understandable (because `OptHop` let think you need to first `OptHop` then `OptApplyAcceptor` wich is false).

Jorropo:
If you want keep OptHop as OptHop I can without a problem, a note just do it.

Jorropo:
@vyzo done, its now truly non breaking (except for RelayOpt type but I can't do much about that, an switch type here will be too costy for that and ugly).

vyzo:
We need to change the `CanHop` handler to apply the filter; we can't have `CanHop` return `OK` and then attempting to hop failing because of the filter application.
Also note that we are planning passive discovery of relays for autorelay, so having `CanHop` return true without applying the filter will cause problems.

Jorropo:
@vyzo its done, see https://github.com/libp2p/go-libp2p-circuit/blob/88de6e6b1bab24289058b6ff8e87b3810c9f4cde/relay.go#L452

relay.go
vyzo: This is not a breaking change, we are adding an error code (and we can update the spec for it).
We do _need_ this error code or else debugging will be a nightmare.
Basically, it's `CANT_SPEAK_REALY` if `r.hop` is false and `RELAY_REFUSED` if the filter rejects the request.

Jorropo: Ok I'm doing this, I didn't though about the fact a new error will code will still be understand as an error even if not known.

Jorropo: @vyzo When talk about `RELAY_REFUSED` are you talking about `STOP_RELAY_REFUSED` or a new `HOP_RELAY_REFUSED` ?

vyzo: `HOP_RELAY_REFUSED`.

cfg/peer-filter/filter.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type PeerFilter struct {
+	// Allowed store who can hop.
+	allowed map[peer.ID]struct{}
+	// An drwmutex is used here because even if drwmutex is slower in monocore
+	// we expect very few Lock but a lot of RLock on a lot of core.
+	dmx ds.DRWMutex
```
vyzo: please use a `sync.RWLock`, we don't need this dependency.

cfg/peer-filter/filter.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	ds "github.com/jonhoo/drwmutex"
+
+	"github.com/libp2p/go-libp2p-circuit"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+)
+
+type PeerFilter struct {
```
vyzo: It would be nice to have a filter that checks network subnets as well, but that's good enough for now.

cfg/peer-filter/filter.go
```diff
@@ -0,0 +1,63 @@
...
+		dmx: ds.New(),
+	}
+
+	for _, p := range ids {
+		pf.allowed[p] = struct{}{}
+	}
+	return &pf
+}
+
+// Allow, allow a peer to hop.
```
vyzo: "Allow allows a peer to hop"

relay.go
```diff
@@ -54,29 +54,20 @@ type Relay struct {
...
-	OptHop = RelayOpt(1)
-	// OptDiscovery configures this relay transport to discover new relays
-	// by probing every new peer. You almost _certainly_ don't want to
-	// enable this.
-	OptDiscovery = RelayOpt(2)
-)
+// Acceptor return true if this conn is allowed to hop on this peer.
+// Its a `network.Stream` and not `peer.ID` because we excpect user to maybe
+// filter by by subnet or by transport.
+type Acceptor func(network.Stream) bool
```
vyzo: We discussed this with @stebalien and we think we also need the target peer as well.
So the filter should be `func(network.Stream, peer.ID) bool`.

vyzo:
So to clarify the request to change the acceptor filter to take the target peer as well.
What we really want is a filter that allows only certain peers (or subnets and whatnot) to use the relay.
This can only be accomplished by having the filter take the target peer -- this is the peer that is going to receive the stop stream, and in general we want to allow connectivity from all peers through the relay (unless the filter blacklists some incoming peer).
As it stands now, it is all backwards. We only check if the inbound peer is allowed to open a hop stream, while what we really want is to check whether the outbound peer is allowed to use the relay to receive stop streams.

vyzo:
It might also make sense to split the filter in two: one for `CanHop` and one for `HopStream`.

relay.go
Jorropo: I think so, I just didn't put because we can get it by `s.Conn().RemotePeer()` and some filter (like subnet didn't require it), including this in the filter allow to do `s.Conn().RemotePeer()` or not depending of the need (reducing memory footprint).

vyzo:
travis complains, can you `go fmt`?

Jorropo:
@vyzo done!

cfg/peer-filter/filter.go
```diff
@@ -0,0 +1,80 @@
...
+func (pf *PeerFilter) IsAllowed(p peer.ID) bool {
+	pf.mx.RLock()
+	_, is := pf.allowed[p]
+	pf.mx.RUnlock()
+	return is
+}
+
+// Used by the relay.
+func (pf *PeerFilter) In(s network.Stream) bool {
+	return pf.IsAllowed(s.Conn().RemotePeer())
```
vyzo: That will kill the functionality, we want the peer filter to allow inbound connections from _all_ peers, not the same set of peers that we allow hop for.

cfg/peer-filter/filter.go
Jorropo: Wrong comment sorry, its only used by utils, this allow the user to create filter where the In and Out filter are not from the same filter (example, if you want to create a filter filtering inbound with subnet and outbound with peer id, you can).

cfg/peer-filter/filter.go
Jorropo: Wrong comment sorry, its only used by utils, this allow the user to create filter where the In and Out filter are not from the same filter (example, if you want to create a filter filtering inbound with subnet and outbound with peer id, you can).

Jorropo:
@vyzo I think I don't understand correctly what `handleHopStream` does, is this used by peer who want to start a relaying session (listening on the relay) or by peer who want to connect to peer with already a listening session ?

vyzo:
this has been superseded by the ACL functionality in v2 relays, so I am closing it.

PR #101, opened 2020-03-31
feat: remove relay discovery and unspecified relay dialing

Stebalien:
This was a misfeature that looks great in demos but isn't useful (and wastes bandwidth and could potentially cause other DoS issues) in practice.

First, random relays we discover are unlikely to support _active_ relaying. This means that picking a random relay is usually going to fail because the random relay won't already be connected to the target.

Second, there are two use-cases for relaying:

1. Connecting to undialable nodes.
2. Relaying all outbound traffic.

For the first case, we only want to use the relays specified by the target peers. For the second case, we'd need to modify either libp2p's dialer logic (in the swarm) to prefix all addresses with the specified relay.

The current logic covers neither use-case so I'm removing it.


Stebalien:
Why?

vyzo:
It is useful for driving relay discovery (say for autorelay); couple that with a relay filters, and we can have a very simple mechanism for setting up relay forwarding.

Stebalien:
We don't use this for AutoRelay, we don't even expose the relays we find. AutoRelay now discovers its own relays via identify events.

Stebalien:
(sorry, I just like deleting code)
KILL KILL KILL!

PR #121, opened 2020-06-24
chore: update deps

Stebalien:
Updates deps to go-libp2p-core 0.6.0.


Stebalien:
Primary motivation: remove unnecessary dependencies from the dependency tree.

PR #124, opened 2020-12-17
Timeout Stream Read

aarshkshah1992:
@vyzo @Stebalien 

I read through the circuit stream handling code and looks like this is the ONLY change we need to make.
Is there anything/any other incoming stream read we need to time out here ?

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -35,6 +35,8 @@ var (
 
 	HopStreamBufferSize = 4096
 	HopStreamLimit      = 1 << 19 // 512K hops for 1M goroutines
+
+	streamTimeout = 1 * time.Minute
 )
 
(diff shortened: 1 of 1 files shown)


vyzo (approved):
LGTM, I don't think we need timeouts anywhere else.

vyzo:
what's with the test failures?

vyzo:
```
ERROR: some files not gofmt'ed:

./relay.go
```

you need to gofmt; can you set your editor to do it automatically?

Stebalien (approved):

aarshkshah1992:
@vyzo My new setup didn't have go fmt configured. Have fixed it.

PR #125, opened 2021-02-04
p2p-circuit v2

vyzo:
v2 of the protocol and Limited Relays.

Depends on:
- https://github.com/libp2p/go-libp2p-core/pull/175
- https://github.com/libp2p/go-libp2p-swarm/pull/236
- https://github.com/libp2p/go-libp2p-transport-upgrader/pull/71

TBD:
- [ ] [spec](https://github.com/libp2p/specs/pull/325) (duh!)
- [x] reservation vouchers
- [x] godocs
- [X] unit tests
- [x] Update and propagate released versions of deps.
- [x] test in the wild

----------------------------------------------------------
# An Overview of the p2p-circuit/v2 Protocol

The relay protocol has been split into two protcols: the Hop protocol and the Stop protocol.
- the Hop protocol governs communication with the relay for reservations and circuit connection initiation.
- the Stop protocol governs termination of circuit connections.

## Establishing a relayed connection

Let's suppose that a peer A wants to connect to peer B through relay peer R.

### Reservations

 Peer B must first establish a reservation by connecting (and maintaining a connection) to R. 

This is done using the Hop protocol and sending a Reserve message. The relay responds with OK if the reservation is accepted and has a specific TTL, after which the reservation will dissipate.

Peer B is responsible for refreshing its reservation when it reaches the end of the TTL.
In the relay logic, reservations are governed by strict resource limits and ACLs.

### Connection Establishment

Peer A connects to Relay R, opens a Hop protocol stream and sends a CONNECT message.

Relay R must have a reservation and connection to peer B. Active relay has been deprecated in the code, but the protocol allows for implementation without any change in the wire protocol.

Once R receives the CONNECT message, it opens a Stop protocol stream to B and sends a CONNECT message.
Peer B responds with OK; the relay R responds with OK to peer A and bridges the two streams in order to establish the relayed connection.

### Connection Limits

Relays are allowed (and encouraged! the relay implementation sets strict limits by default) to limit the relayed connection, by time and bytes relayed in each direction.
The limits are conveyed to both sides of a relayed connection; relayed connections are marked as transient by the transport.

## Implementation Details

Implementation lives under `v2`.

The Stop protocol implementation lives in `v2/client` and provides a Transport suitable for attachment to a host. The implementation mounts the v1 protocol and provides backwards compatibility in dialing v1 peers and accepting relayed connections from v1 peers through v1 relays.

The Hop protocol implementation lives in `v2/relay` and provides a Relay object that can be instantiated on deman. When instantiated, the Relay registers a stream handler from the v2 hop protocol. 

The two are completely independent. The client package implements the transport and is the point of integration with `go-libp2p`. The relay package is the actual relay and can be used in dedicated relays and (as planned for Project Flare) public DHT servers. The default resource limits are very much constrained which makes it safe for public nodes.


vyzo:
Summoning @Stebalien @raulk @yusefnapora @aarshkshah1992 -- this is ready for (initial) review!

vyzo:
Added tests and godocs.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+		log.Debugf("relay copy error: %s", err)
+		// Reset both.
+		src.Reset()
+		dest.Reset()
+	} else {
+		// propagate the close
+		dest.CloseWrite()
+		if count == limit {
+			// we've reached the limit, discard further input
+			src.CloseRead()
```
aarshkshah1992, line 343: _For my own understanding_

 I understand this is nice to have here but are there any repercussions if we don't do this here ?

aarshkshah1992: I mean, since the Limited Reader has already given us an EOF and we're not reading anymore, what's the point of "discarding further input" ?

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+func (r *Relay) disconnected(n network.Network, c network.Conn) {
+	p := c.RemotePeer()
+	if n.Connectedness(p) == network.Connected {
+		return
+	}
+
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	delete(r.rsvp, p)
```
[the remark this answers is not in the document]
aarshkshah1992: Ah, I see we do it as part of the GC. Can probably ignore the above comment.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+	r.mx.Unlock()
+
+	log.Debugf("reserving relay slot for %s", p)
+
+	err := r.writeResponse(s, pbv2.Status_OK, r.makeReservationMsg(p), r.makeLimitMsg(p))
+	if err != nil {
+		s.Reset()
+		log.Debugf("error writing reservation response; retracting reservation for %s", p)
+		r.mx.Lock()
+		delete(r.rsvp, p)
```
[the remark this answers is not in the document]
aarshkshah1992: Nvm, I see we do this in GC and it makes more sense to do it there in case the peer tries connecting to us again before the refresh interval has elapsed. This is smart.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+
+	buf := pool.Get(r.rc.BufferSize)
+	defer pool.Put(buf)
+
+	limitedSrc := io.LimitReader(src, limit)
+
+	count, err := io.CopyBuffer(dest, limitedSrc, buf)
+	if err != nil {
+		log.Debugf("relay copy error: %s", err)
+		// Reset both.
```
aarshkshah1992, line 335: Does this mean that the calls to `stream.Close()` in the `done` function are helpful ONLY if both sides write an EOF to the stream when they are done ?

v2/client/client.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	logging "github.com/ipfs/go-log"
+	tptu "github.com/libp2p/go-libp2p-transport-upgrader"
+)
+
+const (
+	ProtoIDv1     = "/libp2p/circuit/relay/0.1.0"
```
aarshkshah1992, line 15: de-duplicate all 3.

aarshkshah1992 (commented):
@vyzo 

Did a first pass on the Relay code. Will move to the client now.

Awesome stuff man !

v2/pb/circuit.proto
```diff
@@ -0,0 +1,61 @@
+syntax = "proto2";
+
+package circuit.pb;
```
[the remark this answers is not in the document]
vyzo: that's the job of the spec!

v2/util/io.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/multiformats/go-varint"
+)
+
+type DelimitedReader struct {
+	r   io.Reader
+	buf []byte
+}
+
+// The gogo protobuf NewDelimitedReader is buffered, which may eat up stream data.
```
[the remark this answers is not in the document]
vyzo: old problem, the buffer can and will eat stream data send immediately after the handshake if they are in the same packet; it bit us in relay v1.

v2/util/io.go
```diff
@@ -0,0 +1,67 @@
+package util
+
```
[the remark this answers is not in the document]
vyzo: meh.... if I have nothing better to do.

v2/relay/acl.go
```diff
@@ -0,0 +1,17 @@
+package relay
+
+import (
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// ACLFilter is an Access Control mechanism for relayed connect.
+type ACLFilter interface {
```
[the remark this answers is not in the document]
vyzo: I don't think so, it's very relay specific.

v2/relay/acl.go
```diff
@@ -0,0 +1,17 @@
...
+	// AllowReserve returns true if a reservation from peer a given peer with a given multiaddr
```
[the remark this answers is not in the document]
vyzo: right, will fix.

v2/relay/resources.go
```diff
@@ -0,0 +1,58 @@
...
+	BufferSize int
+}
+
+// RelayLimit are the per relayed connection resource limits.
+type RelayLimit struct {
+	// Duration is the time limit before resetting a relayed connection; defaults to 1min.
+	Duration time.Duration
+	// Data is the limit of data relayed (on each direction) before resetting the connection.
+	// Defaults to 64KB
+	Data int64
```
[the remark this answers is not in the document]
vyzo: It's about 200 bytes, not an issue.  I will increase the default to 128K regardless.

v2/relay/resources.go
```diff
@@ -0,0 +1,58 @@
...
+	// sooner than that, then the refresh will be refused.
+	// Defaults to 15min.
+	ReservationRefreshTTL time.Duration
+
+	// MaxReservations is the maximum number of active relay slots; defaults to 128.
+	MaxReservations int
+	// MaxCircuits is the maximum number of open relay connections for each peer; defaults to 16.
+	MaxCircuits int
+	// BufferSize is the size of the relayed connection buffers; defaults to 2048.
+	BufferSize int
```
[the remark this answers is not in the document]
vyzo: yes, that's buffer size -- important for controlling memory use.
I think the name is fine, we don't have to be overly verbose.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+
+func (r *Relay) handleStream(s network.Stream) {
+	select {
+	case <-r.ctx.Done():
+		s.Reset()
+		return
+	default:
+	}
+
+	s.SetReadDeadline(time.Now().Add(StreamTimeout))
```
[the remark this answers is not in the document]
vyzo: eh? that doesn't do anything.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+	select {
+	case <-r.ctx.Done():
+	default:
+		r.cancel()
+	}
+	return nil
+}
+
+func (r *Relay) handleStream(s network.Stream) {
+	select {
```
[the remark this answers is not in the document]
vyzo: Oh yeah, didn't realize we even had an interface to remove the stream handler; will do.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+
+	case pbv2.HopMessage_CONNECT:
+		r.handleConnect(s, &msg)
+
+	default:
+		r.handleError(s, pbv2.Status_MALFORMED_MESSAGE)
+	}
+}
+
+func (r *Relay) handleReserve(s network.Stream, msg *pbv2.HopMessage) {
```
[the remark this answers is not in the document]
vyzo: cosmetics....

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+	// TODO cache this
+	ai := peer.AddrInfo{r.host.ID(), r.host.Addrs()}
+
+	return &pbv2.Reservation{
+		Ttl:   &ttl,
+		Relay: util.PeerInfoToPeerV2(ai),
+	}
+}
+
+func (r *Relay) makeLimitMsg(p peer.ID) *pbv2.Limit {
```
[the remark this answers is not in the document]
vyzo: for now, it can become useful for establishing per peer limits in the future (eg peer is allowed full unlimited relay).

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+	var msg pbv2.HopMessage
+	msg.Type = pbv2.HopMessage_STATUS.Enum()
+	msg.Status = status.Enum()
+	msg.Reservation = rsvp
+	msg.Limit = limit
+
+	return wr.WriteMsg(&msg)
+}
+
+func (r *Relay) makeReservationMsg(p peer.ID) *pbv2.Reservation {
```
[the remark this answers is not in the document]
vyzo: for now -- it will become necessary when we make reservation records, as those will need to embed the peer.

v2/relay/relay.go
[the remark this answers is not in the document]
vyzo: refresh logic has been removed.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	pool "github.com/libp2p/go-buffer-pool"
+
+	logging "github.com/ipfs/go-log"
+)
+
+const (
+	ProtoIDv2Hop  = "/libp2p/circuit/relay/0.2.0/hop"
+	ProtoIDv2Stop = "/libp2p/circuit/relay/0.2.0/stop"
```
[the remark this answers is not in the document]
vyzo: well, small potatoes :)

v2/relay/relay.go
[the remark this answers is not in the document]
vyzo: refresh has been removed.

v2/relay/relay.go
vyzo: we'll keep buffering data, so it's a must do.

v2/relay/relay.go
vyzo: yes, it needs to be there.

v2/client/client.go
vyzo: meh.... maybe.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+	acl  ACLFilter
+
+	mx      sync.Mutex
+	rsvp    map[peer.ID]time.Time
+	refresh map[peer.ID]time.Time
+	conns   map[peer.ID]int
+}
+
+// New constructs a new limited relay that can provide relay services in the given host.
+func New(ctx context.Context, h host.Host, opts ...Option) (*Relay, error) {
```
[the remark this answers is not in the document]
vyzo: sure, easy to do. I had it like this in the beginning, but then added the context.

v2/relay/resources.go
```diff
@@ -0,0 +1,58 @@
...
+	// ReservationRefreshTTL is the time that must have ellapsed since the last reservation
+	// in order to accept a new reservation. If a client attempts to refresh a reservation
+	// sooner than that, then the refresh will be refused.
+	// Defaults to 15min.
+	ReservationRefreshTTL time.Duration
+
+	// MaxReservations is the maximum number of active relay slots; defaults to 128.
+	MaxReservations int
+	// MaxCircuits is the maximum number of open relay connections for each peer; defaults to 16.
+	MaxCircuits int
```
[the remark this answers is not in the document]
vyzo: nah, let's not be too verbose.

v2/relay/resources.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Resources are the resource limits associated with the relay service.
+type Resources struct {
+	// Limit is the (optional) relayed connection limits.
+	Limit *RelayLimit
+
+	// ReservationTTL is the duration of a new (or refreshed reservation).
+	// Defaults to 1hr.
+	ReservationTTL time.Duration
+	// ReservationRefreshTTL is the time that must have ellapsed since the last reservation
```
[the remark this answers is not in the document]
vyzo: refresh has been removed.

v2/relay/relay.go
```diff
@@ -0,0 +1,483 @@
...
+
+	log.Debugf("reserving relay slot for %s", p)
+
+	err := r.writeResponse(s, pbv2.Status_OK, r.makeReservationMsg(p), r.makeLimitMsg(p))
+	if err != nil {
+		s.Reset()
+		log.Debugf("error writing reservation response; retracting reservation for %s", p)
+		r.mx.Lock()
+		delete(r.rsvp, p)
+		r.host.ConnManager().UntagPeer(p, "relay-reservation")
```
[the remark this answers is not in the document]
vyzo: actually we do, because the peer might disconnect in the meantime and result into an untagging first which would stick an early tag.

v2/client/reservation.go
```diff
@@ -0,0 +1,95 @@
...
+	msg.Type = pbv2.HopMessage_RESERVE.Enum()
+
+	if err := wr.WriteMsg(&msg); err != nil {
+		s.Reset()
+		return nil, fmt.Errorf("error writing reservation message: %w", err)
+	}
+
+	msg.Reset()
+
+	if err := rd.ReadMsg(&msg); err != nil {
```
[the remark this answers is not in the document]
aarshkshah1992: maybe set a deadline for the whole thing.

v2/client/handlers.go
```diff
@@ -0,0 +1,170 @@
...
+		writeResponse: func() error {
+			return writeResponse(pbv2.Status_OK)
+		},
+	}:
+	case <-time.After(AcceptTimeout):
+		handleError(pbv2.Status_CONNECTION_FAILED)
+	}
+}
+
+func (c *Client) handleStreamV1(s network.Stream) {
```
aarshkshah1992, line 90: Is this so that a v2 Stop peer can accept Relayed connections from a v1 Hop peer ?

v2/client/listen.go
```diff
@@ -0,0 +1,54 @@
...
+		Remote: "any",
+	}
+}
+
+func (l *Listener) Multiaddr() ma.Multiaddr {
+	return circuitAddr
+}
+
+func (l *Listener) Close() error {
+	// noop for now
```
aarshkshah1992, line 52: Is this because we currently  don't have an API for/ never stop listening on a particular address in libp2p ?

v2/client/dial.go
```diff
@@ -0,0 +1,161 @@
...
+
+	err := wr.WriteMsg(&msg)
+	if err != nil {
+		s.Reset()
+		return nil, err
+	}
+
+	msg.Reset()
+
+	err = rd.ReadMsg(&msg)
```
aarshkshah1992, line 96: deadline

v2/client/dial.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	status := msg.GetStatus()
+	if status != pbv2.Status_OK {
+		s.Reset()
+		return nil, fmt.Errorf("error opening relay circuit: %s (%d)", pbv2.Status_name[int32(status)], status)
+	}
+
+	var stat network.Stat
+	if limit := msg.GetLimit(); limit != nil {
```
aarshkshah1992, line 114: Please can we document why we do all this ?

v2/client/dial.go
```diff
@@ -0,0 +1,161 @@
...
+		stat.Transient = true
+		stat.Extra = make(map[interface{}]interface{})
+		stat.Extra[StatLimitDuration] = time.Duration(limit.GetDuration()) * time.Second
+		stat.Extra[StatLimitData] = limit.GetData()
+	}
+
+	return &Conn{stream: s, remote: dest, stat: stat, client: c}, nil
+}
+
+func (c *Client) connectV1(s network.Stream, dest peer.AddrInfo) (*Conn, error) {
```
aarshkshah1992, line 124: This is so that a V2 client can dial a V1 Hop.

v2/client/handlers.go
```diff
@@ -0,0 +1,170 @@
+package client
+
```
aarshkshah1992, line 2: Please can we document why we need handlers for V2 & V1 STOP streams ?

v2/client/dial.go
```diff
@@ -0,0 +1,161 @@
+package client
```
aarshkshah1992, line 1: Please can we document why need both connect V1 & V2 ?

aarshkshah1992 (commented):
@vyzo Finished first pass of client too.

Will give all this one more in-depth look tomorrow. But looking amazing so far.

v2/relay/relay.go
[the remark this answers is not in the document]
aarshkshah1992: @vyzo Yeah, you just gotta clock on the "Accept" button man :)

v2/client/conn.go
```diff
@@ -0,0 +1,139 @@
...
+
+// ConnStat interface
+var _ network.ConnStat = (*Conn)(nil)
+
+func (c *Conn) Stat() network.Stat {
+	return c.stat
+}
+
+// conn manager hop tagging
+func (c *Conn) tagHop() {
```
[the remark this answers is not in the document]
vyzo: uhm ok.

v2/relay/relay.go
```diff
@@ -0,0 +1,463 @@
...
+
+	for p, count := range r.conns {
+		if count == 0 {
+			delete(r.conns, p)
+		}
+	}
+}
+
+func (r *Relay) disconnected(n network.Network, c network.Conn) {
+	p := c.RemotePeer()
```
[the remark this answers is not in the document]
vyzo: it's fine.

v2/client/transport.go
```diff
@@ -0,0 +1,80 @@
...
+	return c.upgrader.UpgradeOutbound(ctx, c, conn, p)
+}
+
+func (c *Client) CanDial(addr ma.Multiaddr) bool {
+	_, err := addr.ValueForProtocol(ma.P_CIRCUIT)
+	return err == nil
+}
+
+func (c *Client) Listen(addr ma.Multiaddr) (transport.Listener, error) {
+	// TODO connect to the relay and reserve slot if specified
```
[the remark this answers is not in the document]
vyzo: that's a different topic -- it is for explicit relay address listening, when/if we ever support it.

v2/client/reservation.go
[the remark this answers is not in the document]
vyzo: will do.

v2/client/conn.go
```diff
@@ -0,0 +1,139 @@
...
+func (c *Conn) SetReadDeadline(t time.Time) error {
+	return c.stream.SetReadDeadline(t)
+}
+
+func (c *Conn) SetWriteDeadline(t time.Time) error {
+	return c.stream.SetWriteDeadline(t)
+}
+
+// TODO: is it okay to cast c.Conn().RemotePeer() into a multiaddr? might be "user input"
+func (c *Conn) RemoteMultiaddr() ma.Multiaddr {
```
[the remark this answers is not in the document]
vyzo: ugh.... maybe, I am not inclined to do it right now.

v2/client/reservation.go
```diff
@@ -0,0 +1,95 @@
...
+	// LimitData is the number of bytes that the relay will relay in each direction before
+	// resetting a relayed connection.
+	LimitData int64
+
+	// TODO reservation voucher
+}
+
+// Reserve reserves a slot in a relay and returns the reservation information.
+// Clients must reserve slots in order for the relay to relay connections to them.
+func Reserve(ctx context.Context, h host.Host, ai peer.AddrInfo) (*Reservation, error) {
```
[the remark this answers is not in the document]
vyzo: It will make use of the Expiration/TTL so that it knows when to refresh.
The others may or may not be used, so why not include them? For instance, if those are missing then it's an unlimited relay, which is useful to know.

v2/client/handlers.go
vyzo: yes.

v2/client/listen.go
vyzo: yes.

v2/client/dial.go
vyzo: will do.

v2/client/dial.go
vyzo: uhm ok.

v2/client/dial.go
vyzo: yes.

v2/client/handlers.go
vyzo: It's obviously for backwards compatibility and drop in replacement, but I will add some documentation in the client struct; it doesn't belong here.

v2/client/dial.go
vyzo: see above.

v2/util/io.go
```diff
@@ -0,0 +1,67 @@
...
+// So we need to implement a compatible delimited reader that reads unbuffered.
+// There is a slowdown from unbuffered reading: when reading the message
+// it can take multiple single byte Reads to read the length and another Read
+// to read the message payload.
+// However, this is not critical performance degradation as
+// - the reader is utilized to read one (dialer, stop) or two messages (hop) during
+//   the handshake, so it's a drop in the water for the connection lifetime.
+// - messages are small (max 4k) and the length fits in a couple of bytes,
+//   so overall we have at most three reads per message.
+func NewDelimitedReader(r io.Reader, maxSize int) *DelimitedReader {
```
aarshkshah1992, line 29: need to ensure `maxSize` is gt 0 here, otherwise `ReadByte` can panic later on.

v2/util/pbconv.go
```diff
@@ -0,0 +1,80 @@
...
+	if p == nil {
+		return peer.AddrInfo{}, errors.New("nil peer")
+	}
+
+	id, err := peer.IDFromBytes(p.Id)
+	if err != nil {
+		return peer.AddrInfo{}, err
+	}
+
+	var addrs []ma.Multiaddr
```
aarshkshah1992, line 24: addrs := make([]ma.Multiaddr, 0 , len(p.Addrs )) for allocs.

v2/util/pbconv.go
```diff
@@ -0,0 +1,80 @@
...
+		if err == nil {
+			addrs = append(addrs, a)
+		}
+	}
+
+	return peer.AddrInfo{ID: id, Addrs: addrs}, nil
+}
+
+func PeerInfoToPeerV1(pi peer.AddrInfo) *pbv1.CircuitRelay_Peer {
+	var addrs [][]byte
```
aarshkshah1992, line 36: use make since length is known

v2/util/pbconv.go
```diff
@@ -0,0 +1,80 @@
...
+	if p == nil {
+		return peer.AddrInfo{}, errors.New("nil peer")
+	}
+
+	id, err := peer.IDFromBytes(p.Id)
+	if err != nil {
+		return peer.AddrInfo{}, err
+	}
+
+	var addrs []ma.Multiaddr
```
aarshkshah1992, line 58: make

v2/util/pbconv.go
```diff
@@ -0,0 +1,80 @@
...
+		if err == nil {
+			addrs = append(addrs, a)
+		}
+	}
+
+	return peer.AddrInfo{ID: id, Addrs: addrs}, nil
+}
+
+func PeerInfoToPeerV2(pi peer.AddrInfo) *pbv2.Peer {
+	var addrs [][]byte
```
aarshkshah1992, line 70: make

aarshkshah1992 (commented):
some small changes.

v2/util/io.go
vyzo: Sure, but I wouldn't dwell on that, we are the only users of this.

v2/util/pbconv.go
vyzo: this is a micro-optimization that will do more harm than good -- this is 0-sized almost certainly and we want to keep that nil around.

v2/util/pbconv.go
vyzo: again, same thing -- it doesn't gain anything.

v2/util/pbconv.go
vyzo: I will humor you however and add that code just so you are happy :)

v2/relay/relay.go
```diff
@@ -0,0 +1,463 @@
...
+	h.Network().Notify(
+		&network.NotifyBundle{
+			DisconnectedF: r.disconnected,
+		})
+	go r.background()
+
+	return r, nil
+}
+
+func (r *Relay) Close() error {
```
aarshkshah1992, line 83: Let's cancel the context and wait till the go-routine finishes to ensure that we've wrapped up everything before we return from `Close`.

Best to use a `sync.Once` here to ensure idempotency.

vyzo: we can't wait for everything to finish, there is no good mechanism to do that -- and we might have unlimited relay streams that never finish.

but fine on making it idempotent, an atomic variable will do.

vyzo: so no on making it wait, yes on using a better idempotency mechanism that can support concurrent closes without fear of panic.

v2/relay/relay.go
```diff
@@ -0,0 +1,463 @@
...
+func (r *Relay) gc() {
+	r.mx.Lock()
+	defer r.mx.Unlock()
+
+	now := time.Now()
+
+	for p, expire := range r.rsvp {
+		if expire.Before(now) {
+			delete(r.rsvp, p)
+			r.host.ConnManager().UntagPeer(p, "relay-reservation")
```
aarshkshah1992, line 442: Right, so there's no point in actively pruning the connections we're relaying for this peer(max 16) because wer're untagging it here and are sure these will be pruned later by the Conn Manager ?

v2/relay/relay.go
```diff
@@ -0,0 +1,463 @@
...
+	log.Debugf("relayed %d bytes from %s to %s", count, srcID, destID)
+}
+
+func (r *Relay) relayUnlimited(src, dest network.Stream, srcID, destID peer.ID, done func()) {
+	defer done()
+
+	buf := pool.Get(r.rc.BufferSize)
+	defer pool.Put(buf)
+
+	count, err := io.CopyBuffer(dest, src, buf)
```
aarshkshah1992, line 350: Given that this is essentially a stream read(here and in `relayLimited `), should we be worried about not having set a stream read deadline ?

v2/client/dial.go
```diff
@@ -0,0 +1,169 @@
...
+
+	err := wr.WriteMsg(&msg)
+	if err != nil {
+		s.Reset()
+		return nil, err
+	}
+
+	msg.Reset()
+
+	err = rd.ReadMsg(&msg)
```
aarshkshah1992, line 151: need to set a read deadline for this stream.

aarshkshah1992 (commented):
Finished last round of review.

Will take a look at the tests tomorrow but looks great otherwise.

v2/relay/relay.go
```diff
@@ -0,0 +1,463 @@
...
+
+	done := func() {
+		if atomic.AddInt32(goroutines, -1) == 0 {
+			s.Close()
+			bs.Close()
+			cleanup()
+		}
+	}
+
+	if r.rc.Limit != nil {
```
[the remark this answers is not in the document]
vyzo: that's just not worthed and also the goroutines may *never* finish because it's an unlimited relay.
So, *no*.

v2/relay/relay.go
vyzo: yes, if there is pressure they will be pruned.

v2/relay/relay.go
vyzo: No, this is not relevant for unlimited relay. 
For limited relay, the deadline is set before spawning the goroutine.

v2/client/dial.go
vyzo: sure.

v2/relay/relay.go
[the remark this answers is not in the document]
vyzo: I repeat this, we can't wait. Will just use an atomic to ensure thread-safety and that's about it.

v2/relay/relay.go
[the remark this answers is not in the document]
aarshkshah1992: I am not sure this is the right approach to take. We can easily stop relaying on `unlimited relays` by closing connections to all peers we have opened circuits for.

When `Close` returns, the caller expects that the subsystem has cleaned-up after it itself. Not doing this can leak go-routines if the caller dosen't know this implementation detail.

cc @Stebalien Isn't this in-line with what `Close` does for other sub-systems in libp2p ?

v2/client/reservation.go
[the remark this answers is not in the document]
aarshkshah1992: Make sense.

v2/relay/relay.go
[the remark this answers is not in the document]
vyzo: If you insist on closing the connections it is much easier to tag the Hop/Stop streams and go and reset them; there is no tracking needed and it is much easier to implement (and doesn't suffer from the problem of unlimited relay conns which will never exit).

Waiting for the goroutines is irrelevant.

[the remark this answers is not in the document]
vyzo: Actually, we don't even have to tag streams, which could be a little problematic with the current APIs; we can just look for incoming Hop streams and outgoing Stop streams and reset them.

v2/relay/ipcs.go
```diff
@@ -0,0 +1,110 @@
...
+
+	asnutil "github.com/libp2p/go-libp2p-asn-util"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+var (
+	ErrNoIP              = errors.New("no IP address associated with peer")
+	ErrTooManyPeersInIP  = errors.New("too many peers in IP address")
+	ErrTooManyPeersInASN = errors.New("too many peers in ASN")
```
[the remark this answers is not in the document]
vyzo: Yeah, the only reason I exported them is so that the test could could live together with the other tests in the `test` directory.
It felt odd to have just one test in here.

v2/client/client.go
vyzo: done.

v2/relay/relay.go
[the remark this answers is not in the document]
vyzo: deduplicated.

v2/client/reservation.go
```diff
@@ -95,6 +96,23 @@ func Reserve(ctx context.Context, h host.Host, ai peer.AddrInfo) (*Reservation,
 		result.Addrs = append(result.Addrs, a)
 	}
 
+	voucherBytes := rsvp.GetVoucher()
```
aarshkshah1992, line 99: @vyzo We also need code in the v2 dialer that validates a voucher before dialing.

vyzo: not yet, it's *very* premature -- first we have to figure out how to distribute the vouchers.

v2/pb/circuit.proto
```diff
@@ -0,0 +1,60 @@
...
+}
+
+message Limit {
+  optional int32 duration = 1; // seconds
+  optional int64 data = 2;     // bytes
+}
+
+enum Status {
+  OK                      = 100;
+  RESERVATION_REFUSED     = 200;
```
mxinden, line 53: How about splitting `Status` into a `HopMessage` specific `Status` and a `StopMessage` specific `Status` like done with `Type`? The split would enforce protocol semantics at the datastructure level. E.g. it encodes the semantic that one could not receive a `RESERVATION_REFUSED` when sending a `StopMessage`.

vyzo: We could I guess, but there would be a lot of duplication; I don't feel strong about it nonetheless.

Stebalien:
> Relay R must have a reservation and connection to peer B. Active relay has been deprecated in the code, but the protocol allows for implementation without any change in the wire protocol.

Can we include a flag in the reservation response that specifies whether or not the relay supports active relays? That way the peer can disconnect if it sees that active relay is supported.

We can also just punt this and do it later as it should be backwards compatible.

v2/proto/voucher.go
```diff
@@ -0,0 +1,139 @@
...
+
+type ReservationVoucher struct {
+	// Relay is the ID of the peer providing relay service
+	Relay peer.ID
+	// Peer is the ID of the peer receiving relay service through Relay
+	Peer peer.ID
+	// Expiration is the expiration time of the reservation
+	Expiration time.Time
+	// Signature is the signature of this voucher, as produced by the Relay peer
+	Signature []byte
```
Stebalien, line 21: Can we just use our signed envelopes?

Stebalien: Also, I'm _pretty_ sure we need the relays key for this to be useful. The envelope covers that, IIRC.

v2/proto/voucher.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (rv *ReservationVoucher) bytes() []byte {
+	buf := make([]byte, 1024)
```
Stebalien, line 25: :angry:

v2/proto/voucher.go
```diff
@@ -0,0 +1,139 @@
...
+	relayBytes := []byte(rv.Relay)
+	peerBytes := []byte(rv.Peer)
```
Stebalien, line 27: nit: we could avoid an allocation here if we just copied directly from the strings (which we can do).

v2/proto/voucher.go
```diff
@@ -0,0 +1,139 @@
...
+	Relay peer.ID
+	// Peer is the ID of the peer receiving relay service through Relay
+	Peer peer.ID
+	// Expiration is the expiration time of the reservation
+	Expiration time.Time
+	// Signature is the signature of this voucher, as produced by the Relay peer
+	Signature []byte
+}
+
+func (rv *ReservationVoucher) bytes() []byte {
```
Stebalien, line 24: Why not just use a protobuf?

Stebalien: Or CBOR?

v2/proto/voucher.go
Stebalien: Either that or we have to connect to the relay to get its key, which kind of makes the whole voucher thing pointless (at that point, we might as well just ask the relay "you relaying for peer X?").

v2/proto/voucher.go
```diff
@@ -0,0 +1,139 @@
...
+
+	blob := rv.bytes()
+	result := make([]byte, len(blob)+len(rv.Signature))
+	copy(result, blob)
+	copy(result[len(blob):], rv.Signature)
+
+	return result, nil
+}
+
+func (rv *ReservationVoucher) Unmarshal(blob []byte) error {
```
Stebalien, line 86: Yeah, let's _really_ not do this ourselves.

v2/relay/ipcs.go
```diff
@@ -0,0 +1,110 @@
...
+	peersInIP := ipcs.ips[ips]
+	if len(peersInIP) >= ipcs.iplimit {
+		return ErrTooManyPeersInIP
+	}
+
+	var peersInAsn map[peer.ID]struct{}
+	asn, _ := asnutil.Store.AsnForIPv6(ip)
+	peersInAsn = ipcs.asns[asn]
+	if len(peersInAsn) >= ipcs.asnlimit {
+		return ErrTooManyPeersInASN
```
Stebalien, line 61: :+1: (but I can still just repeatedly reconnect)

v2/relay/relay.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+}
+
+func (r *Relay) relayLimited(src, dest network.Stream, srcID, destID peer.ID, limit int64, done func()) {
+	defer done()
+
+	buf := pool.Get(r.rc.BufferSize)
+	defer pool.Put(buf)
+
+	limitedSrc := io.LimitReader(src, limit)
```
Stebalien, line 343: It would be _nice_ if we had some form of rate-limit as well instead of just an absolute limit.

v2/relay/relay.go
```diff
@@ -0,0 +1,507 @@
...
+		log.Debugf("relay copy error: %s", err)
+		// Reset both.
+		src.Reset()
+		dest.Reset()
+	} else {
+		// propagate the close
+		dest.CloseWrite()
+		if count == limit {
+			// we've reached the limit, discard further input
+			src.CloseRead()
```
Stebalien, line 356: We should _reset_ both streams if there's an error here, not close them.

Stebalien (commented):
1. Attacker connects.
2. Attacker reserves.
3. Attacker uses entire budget.
4. Attacker disconnects.
5. Attacker re-rolls their peer ID and starts at 1.

I think we need:

1. A bit of local history (i.e., a per-peer/ip range cooldown period).
2. A global rate-limit.
3. Ideally, some form of "uses less than X bits/s" that can prevent large spikes.

v2/proto/voucher.go
vyzo: Yeah you are right, I will change it to use the envelope framework.

v2/relay/relay.go
vyzo: agreed, although not exactly clear how to do that.

v2/relay/relay.go
vyzo: limit reached is not an error, but yes, we should check if there is one.

v2/relay/relay.go
Stebalien: Discussed offline. The current _limit_ is 128KiB (or something like that). That should be low enough as it is.

v2/relay/relay.go
Stebalien: It's not an error, but we should _still_ reset. If we send a close, it'll look like a clean close. We should only send an EOF if we receive an EOF.

v2/client/reservation.go
```diff
@@ -0,0 +1,123 @@
...
+
+	rsvp := msg.GetReservation()
+	if rsvp == nil {
+		return nil, fmt.Errorf("missing reservation info")
+	}
+
+	result := &Reservation{}
+	result.Expiration = time.Unix(rsvp.GetExpire(), 0)
+
+	for _, ab := range rsvp.GetAddrs() {
```
mxinden, line 90: How about requiring at least one address? As far as I can tell a reservation without a single address of the relay node is useless as one could not announce its relayed address to anyone else, correct?

v2/relay/relay.go
```diff
@@ -0,0 +1,507 @@
...
+func (r *Relay) makeReservationMsg(p peer.ID, expire time.Time) *pbv2.Reservation {
+	expireUnix := expire.Unix()
+
+	var addrBytes [][]byte
+	for _, addr := range r.host.Addrs() {
+		if !manet.IsPublicAddr(addr) {
+			continue
+		}
+
+		raddr := addr.Encapsulate(r.relayAddr)
```
mxinden, line 415: As far as I can tell this adds both `/p2p/<relay-peer-id>` and `/p2p-circuit` to the multiaddr. I would have expected only the former, not the latter in a field called `addrs`. Why not add `/p2p-circuit` on the client side, along with the client address information?

I would deem a Multiaddr postfixed with `/p2p-circuit` useless, given that it misses the client information. A Multiaddr without the `/p2p-circuit` could at least be used to identify the relay node.

What do you think?

vyzo:
Updated for spec compliance with the current DRAFT of the spec.

v2/proto/voucher.go
vyzo: done.

v2/proto/voucher.go
vyzo: protobuf it is.

v2/relay/relay.go
vyzo: done, per spec.

marten-seemann:
Changed the merge target from `master` to `circuit-v2`. Further development will happen in this feature branch.

PR #131, opened 2021-07-06
simplify generation of reservations

marten-seemann:
There's no need for this difficult to reason about rollback procedure. Difficult because it happens after the mutex is unlocked, so other Go routines might already have modified the maps.

Really, the only reason why writing the response would fail is that the stream was closed / reset by the peer. In that case the reservation would not be received by the peer. However, this can also happen for a number of other reasons, e.g. when the stream is reset _after_ the response was written, if the peer disconnects before reading the message, and many others.

### DIFF
--- a/v2/relay/relay.go
+++ b/v2/relay/relay.go
@@ -175,15 +175,12 @@ func (r *Relay) handleReserve(s network.Stream) {
 
 	log.Debugf("reserving relay slot for %s", p)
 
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
I think I get it. Basically, _we_ don't care if the other side doesn't get the requested reservation. In that case, we'll just garbage collect it later.

Can we leave a comment explaining this?

PR #132, opened 2021-07-06
base limits on reservations issued

marten-seemann:
... and don't care about reservations that are removed (and fulfilled).

That means a relay won't _issue_ more than 128 reservations per 30 minute interval, no matter how many of them are used. There are further limits per peer, IP and ASN, that all work the same (limiting the number of reservations per 30 minute interval).


vyzo (commented):
Is this really better? 
We record the reservation before we actually error out and don't remove it then. Doesn't this make it easier to DoS us?

Note: the refactoring of all constraints together is nice and very welcome.

marten-seemann:
I'm not sure I understand how this makes us easier to DoS.

Before this PR, we'd send out reservations (up to a certain limit). Once a reservation has been used up, we don't count it towards our limits any more. If it's not used, it's garbage collected after a while. This, as @Stebalien noted, would allow a node to DoS us by repeatedly making a reservation and immediately using it.
With this PR, we continue counting this reservation, even if it's used. After 30 minutes it is garbage collected, no matter if it has been used or not.

A DoS is only possible if the attacker controls multiple IP addresses (the limit per peer ID doesn't provide any DoS protection, as peer IDs are trivial to generate), as we're limiting the number of reservations to `MaxReservationsPerIP` (default: 4), whereas the total number of reservations is limited to `MaxReservations` (default: 128). Of course, a DDoS would still be easy to pull off, but this applied to the situation before as well, didn't it?

vyzo:
Hrm, you know, this might indeed better than what we have (refactoring aside, which certainly is :)

My concern is that a bunch of peers connect to you, make a reservation each, go over the liimit, and then just disconnect.
That would leave the peer without any reservation slots even though it can serve new peers.

marten-seemann:
> My concern is that a bunch of peers connect to you, make a reservation each, go over the liimit, and then just disconnect.
That would leave the peer without any reservation slots even though it can serve new peers.

Hm. I guess that's by design, isn't it? We limit each relay's utilization to values far below what they _could_ handle, if they're running on decent hardware and with a decent connection. We do this so we can justify turning it on by default.

@Stebalien, wdyt?

v2/relay/resources.go
```diff
@@ -48,8 +51,9 @@ func DefaultResources() Resources {
 		MaxCircuits:     16,
 		BufferSize:      2048,
 
-		MaxReservationsPerIP:  4,
-		MaxReservationsPerASN: 32,
+		MaxReservationsPerPeer: 8,
+		MaxReservationsPerIP:   4,
```
mxinden, line 55: Do I understand correctly that with these limits there can be up to 8 registrations per 30 minutes by a single peer ID and 4 registrations per 30 minutes by a single IP address? If so, why set the former higher than the latter? I would expect multiple peers to use a single IP address (e.g. through NAT), but I doubt many peers use multiple IP addresses.

vyzo: yeah, we should probably reverse these numbers.

marten-seemann: Done.

marten-seemann:
@Stebalien Friendly ping.

v2/relay/resources.go
mxinden: :+1:  Thanks!

Stebalien:
> Hm. I guess that's by design, isn't it? We limit each relay's utilization to values far below what they could handle, if they're running on decent hardware and with a decent connection. We do this so we can justify turning it on by default.

We _could_ oversubscribe by some amount, but that's _probably_ overkill. I'd leave it as-is for now and change it if we notice issues.

Stebalien:
It's always possible to give the hydras a high limit...

v2/relay/constraints.go
```diff
@@ -0,0 +1,187 @@
...
+	total map[uint64]time.Time
+	peers map[peer.ID]map[uint64]time.Time
+	ips   map[string]map[uint64]time.Time
+	asns  map[string]map[uint64]time.Time
+}
+
+// NewConstraints creates a new constraints object.
+// The methods are *not* thread-safe; an external lock must be held if synchronization
+// is required.
+func NewConstraints(rc *Resources) *constraints {
```
Stebalien, line 45: Returning a private type from a public function is generally not the best.

v2/relay/constraints.go
```diff
@@ -0,0 +1,187 @@
...
+	b := make([]byte, 8)
+	crand.Read(b)
```
Stebalien, line 47: (panic on failure)

v2/relay/constraints.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return errors.New("no IP address associated with peer")
+	}
+
+	peerReservations := c.peers[p]
+	if len(peerReservations) >= c.rc.MaxReservationsPerPeer {
+		return errTooManyReservationsForPeer
+	}
+
+	ipStr := ip.String()
```
Stebalien, line 84: can we use `c.ips[string(ip)]`? (avoids some string formatting and an allocation because I'm cheep)

v2/relay/constraints.go
```diff
@@ -0,0 +1,187 @@
...
+		if asn != "" {
+			ansReservations = c.asns[asn]
+			if len(ansReservations) >= c.rc.MaxReservationsPerASN {
+				return errTooManyReservationsForASN
+			}
+		}
+	}
+
+	now := time.Now()
+	id := c.rand.Uint64()
```
Stebalien, line 103: If we need unique IDs, `new(struct{})` should do it (I think, unless you need `new(byte)`?).

Otherwise, I'd just make a `reservation` struct:

```go
type reservation {
    time time.Time
    peer peer.ID
}
```

And make all of these maps `*reservation` sets (`map[*reservation]struct{}`).

v2/relay/constraints.go
```diff
@@ -0,0 +1,187 @@
...
+	ticker := time.NewTicker(cleanupInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-closeChan:
+			return
+		case now := <-ticker.C:
+			c.mutex.Lock()
+			for id, t := range c.total {
+				if t.Add(validity).Before(now) {
```
Stebalien, line 142: nit: Given that we always add validity, maybe just store the expiration instead?

v2/relay/constraints.go
Stebalien: Actually, no, there's a _much_ simpler way to do this. Reservations are increasing so we don't even need a map. We can just keep a sorted list of times.

1. When reserving, append.
2. When cleaning, pop from the beginning.

If we want to get even faster (likely not necessary given the reservation limit) we could keep a queue of reservations to garbage collect (again, in time sorted order). Each entry in the queue would include:

1. The reservation's expiration.
2. The key in the ASN map (if any).
3. The key in the IP map (if any).
4. etc.

This would let us drop the GC loop entirely, garbage collecting old reservations on demand every time we try to add a new reservation.

v2/relay/constraints.go
marten-seemann: Right, there's no need to export the constructor in the first place, now that we moved the tests to this package.

v2/relay/constraints.go
marten-seemann: Fixed in 88d9aab.

v2/relay/constraints.go
marten-seemann: Fixed in e1f7b86.

Stebalien: Ah, sorry, I meant `string(ip)`. That is, convert the IP address to a string directly instead of calling `String()`. That'll avoid formatting and/or allocating (because `mymap[string(foo)]` doesn't allocate on access). Of course, the whole ipv6 versus ipv4 thing might cause some issues?

Otherwise, calling this inline or ahead of time doesn't really matter.

marten-seemann: I try to avoid `string(ip)`, as the `net.IP` type doesn't guarantee that the representation is unique, for example, you can save an IPv4 address in 4 and in 16 bytes. I'm not sure if `manet.ToIP` takes care of that, but even if I don't think we should build code that relies on that property.

v2/relay/constraints.go
marten-seemann: Done.

v2/relay/constraints.go
Stebalien: Fair enough. This likely a pointless micro optimization.

v2/relay/constraints.go
```diff
@@ -0,0 +1,153 @@
...
+	t time.Time
+}
+
+// constraints implements various reservation constraints
+type constraints struct {
+	rc *Resources
+
+	mutex sync.Mutex
+	rand  rand.Rand
+	total *list.List
```
Stebalien, line 37: Any reason not to just use slices?

Stebalien: Because I said lists... sorry. I've been programming in lisp too much lately. A slice is probably much simpler and more performant.

marten-seemann: Yeah, and I've clearly not been thinking clearly. Of course a slice is much better.

v2/relay/constraints.go
```diff
@@ -0,0 +1,135 @@
...
+	errTooManyReservationsForIP   = errors.New("too many peers for IP address")
+	errTooManyReservationsForASN  = errors.New("too many peers for ASN")
+)
+
+// constraints implements various reservation constraints
+type constraints struct {
+	rc *Resources
+
+	mutex sync.Mutex
+	rand  rand.Rand
```
Stebalien, line 31: Do we use this?

v2/relay/constraints.go
```diff
@@ -0,0 +1,135 @@
...
+			break
+		}
+		index = i + 1
+	}
+	return l[index:]
+}
+
+func (c *constraints) cleanup(now time.Time) {
+	c.total = c.cleanupList(c.total, now)
+	for k, peerReservations := range c.peers {
```
Stebalien, line 126: Small optimization: we can return early when nothing has expired (when `c.total` wasn't updated).

Stebalien (approved):

v2/relay/constraints.go
marten-seemann: Nope, and staticcheck didn't catch it because we did initialize it.

v2/relay/constraints.go
marten-seemann: I _think_ you're right, but I'm not sure if it's worth it. This seems like an easy place to introduce non-obvious bugs in the future.

PR #141, opened 2021-09-08
add a Close method, remove the context from the constructor

marten-seemann:
By implementing `io.Closer`, we make sure that the swarm will call `Close` when it's shutting down: https://github.com/libp2p/go-libp2p-swarm/blob/b035b90f5330de123747fe32afb3c2361f3ec0bc/swarm.go#L197-L204

Note that this PR isn't implementing a clean shutdown: it doesn't make sure that all Go routines created by are actually stopped. It's not worth the effort, given that we're going to deprecate this package soon anyway.

### DIFF
--- a/relay.go
+++ b/relay.go
@@ -41,10 +41,11 @@ var (
 
 // Relay is the relay transport and service.
 type Relay struct {
-	host     host.Host
-	upgrader *tptu.Upgrader
-	ctx      context.Context
-	self     peer.ID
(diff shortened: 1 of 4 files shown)


Stebalien:
I'll let @vyzo review first.

vyzo (approved):
It doesn't do much, but I agree let's not cry over spilled milk for soon to be deprecated code.

PR #145, opened 2021-11-10
feat: plumb through contexts from peerstore

guseggert:


marten-seemann (changes requested):
Why is this needed? This PR doesn't do anything but bump versions.

guseggert:
> Why is this needed? This PR doesn't do anything but bump versions.

Basically because I used a script to generate the topological ordering of repos to plumb this through, and considering MVS made it harder than just doing this.  But I can skip this, you are right.

PR #148, opened 2021-12-30
use the resource manager

marten-seemann:


vyzo:
wait, haven't we deprecated this repo? We are not integrating this anywhere, so it is useless to update.

marten-seemann:
Unfortunately, we still use this repo for compat tests in go-libp2p, so we have to do all interfaces changes so that our code continues building.

